Issue #14, opened 2017-12-19
As an admin, I want to add and edit collection descriptions so that I can help site users understand the collection and find related materials.


rlskoeser:
## Notes for testing
- There should be a description on collections
- You can edit it, and also add bolding and italics to fonts
- These will display on the 'View Collection' page with formatting intact.

## Notes for development
- Needs to support for basic formatting
- Suggest tinymce with fairly minimal set of html fields/styles
- make sure description displays in list view with formatting

bwhicks:
Documenting conversation with @xinyil re: the styling editor: I've left on functionality for admin users to bold (`<strong>`), italicize (`em`), and add links, but nothing else in TinyMCE.

meagan1019:
Both superuser and archive manager were able to add collection description, which is view able on the user site with proper formatting (italics, bold, etc.).

PR #21, opened 2018-02-05
Feature/collection descriptions

bwhicks:
Adds functionality for issue #14 (collection descriptions) implemented via Mezzanine's `RichTextField` and customized TinyMCE.


rlskoeser:
@bwhicks thanks for making the changes, looks like a good solution.

Issue #16, opened 2017-12-19
As an admin, I want a way to search and select digitized items for bulk addition to a collection so that I can efficiently organize large groups of items.


rlskoeser:
## Notes for testing
- [x] Select a few digitized works in the admin interface and then select (lower left) from the dropdown to add them to a collection. This will take you to an intermediate page where you can select collections and then finalize the add. 
- [x] On successful add you should be returned to the digitzed works listing with a success message and the collections should be set.
- [x] Try doing this by selecting all items on the page (check box in the top row) and then using the `select all 5007` option in lower left and add everything to a collection. This will take several seconds, but should work.
- [x] Try searching the digitized works for something that returns more than 100 items (e.g. "elocution" returns 259) and add all items from the search to a collection.

## Notes for development
- Implement as a custom django admin action - see https://docs.djangoproject.com/en/2.0/ref/contrib/admin/actions/
- Should provide an intermediate page for user to select the collection - https://docs.djangoproject.com/en/2.0/ref/contrib/admin/actions/#actions-that-provide-intermediate-pages 
- The action shoud be specific to/available for DigitizedWorks only

mnaydan:
I was able to add selected works to one collection, and to multiple collections. I was able to add all 5007 items to a collection. I searched by the keyword "elocution" and was able to add all 259 items to a collection. The bulk addition is therefore functional, but @meagan1019 and I don't know if it's the most useful for our purposes. We will be assigning items to collections primarily by volume ID. Meredith will be going through the admin site or a CSV and assigning collections (sometimes 2-3 per volume) by hand.

rlskoeser:
If you're assigning one or more collections to individual items by hand then doesn't the collection editing provided by #15 meet your needs?

mnaydan:
Yes, that is true. @meagan1019 explained that MM likes to identify collections in an excel file quickly in a column via a shorthand, like [P, H]. Then Meagan and I would have to go in to each item in the admin and assign collections to individual items by hand, which we could do with #15, but it would be time consuming. If we could use Ben's interface to search multiple volume IDs at once, both as a list and as boolean OR [hathiid OR hathiid OR hathiid], even for 100 IDs at a time, that would cut down the time significantly.

rlskoeser:
@mnaydan thanks for explaining.  Let's ask MM to try the collection editing options we've implemented and then find a time to discuss in person.

mnaydan:
[...]

PR #22, opened 2018-02-07
Feature/collection bulkadd

bwhicks:
Adds a Django admin action for bulk adding selected digitized works for collections (#16), as well as associated views and tests.


ppa/archive/tests/test_admin.py
```diff
@@ -46,3 +48,15 @@ class Meta:
...
+        # create a DigitizedWorkAdmin object
+        digworkadmin = DigitizedWorkAdmin(DigitizedWork, AdminSite())
+        fakerequest = Mock()
+        # mock items 1,2,3 being selected
+        fakerequest.POST.getlist.return_value = ['1', '2', '3']
+        redirect = digworkadmin.bulk_add_collection(fakerequest, [])
+        # should return a redirect
+        assert isinstance(redirect, HttpResponseRedirect)
+        # url should reverse the appropriate route and append ?ids=1,2,3
+        assert redirect.url == '%s?ids=1,2,3' % reverse('archive:bulk-add')
```
rlskoeser: Any error cases that should be tested here?

ppa/archive/urls.py
```diff
@@ -1,13 +1,20 @@
...
 from ppa.archive import views
 
 
 urlpatterns = [
     url('^collections/$', views.CollectionListView.as_view(),
         name='list-collections'),
+    url(
+        '^collections/add/$',
+        staff_member_required(views.BulkAddCollectionView.as_view()),
+        name='bulk-add'
```
rlskoeser: I think this url name isn't specific enough and might get confusing as we add more views.  Could you revise to include collection somehow?  Maybe bulk is unnecessary, since we're unlikely to have other add-to-collection views.

The url isn't very resource-y, but I'm having trouble coming up with anything better.

ppa/archive/views.py
```diff
@@ -166,3 +170,73 @@ def render_to_csv(self, data):
...
+    def get_queryset(self, *args, **kwargs):
+        '''Return a queryset filtered by id, or empty list if no ids'''
+        ids = self.request.GET.get('ids', None)
+        # if somehow a problematic non-pk is pushed, will be ignored in filter
+        return DigitizedWork.objects.filter(id__in=ids.split(',')
+                                            if ids else [])
+
+    def get_initial(self):
+        '''
+        Set the hidden field with the :class:ppa.archive.models.DigitizedWork
```
rlskoeser: I'm not clear what you're doing here; please clarify in the docstring and/or comments.  Is this just overriding the default method to filter out any invalid ids?

ppa/archive/views.py
```diff
@@ -166,3 +170,73 @@ def render_to_csv(self, data):
...
+        ids = ','.join(str(val) for val in
+                       self.get_queryset().values_list('id', flat=True))
+        if ids:
+            initial['digitized_work_ids'] = ids
+        return initial
+
+    def post(self, request, *args, **kwargs):
+        '''
+        Add :class:ppa.archive.models.DigitizedWork instances passed in form
+        data to selected instances of :class:ppa.archive.models.Collection,
```
rlskoeser: I think your `:class:` docstring syntax isn't quite right (don't you need backticks around class names?).  You might want to build the sphinx doc and check.

ppa/archive/views.py
```diff
@@ -166,3 +170,73 @@ def render_to_csv(self, data):
...
+            solr_docs = [work.index_data() for work in digitized_works]
+            solr, solr_collection = get_solr_connection()
+            solr.index(solr_collection, solr_docs,
+                params={'commitWithin': 2000})
+            # create a success message to add to message framework stating
+            # what happened
+            htids = ', '.join(work.source_id for work in digitized_works)
+            collections = ', '.join(collection.name for
+                                    collection in data['collections'])
+            messages.success(request, '%s successfully added to: %s.'
```
rlskoeser: I tried testing this with 100 works selected - the form works fine, but the message fails because it is too long.  Maybe just report the total number of works added to which collections?

ppa/archive/admin.py
```diff
@@ -27,6 +31,22 @@ def save_related(self, request, form, formsets, change):
         digwork = DigitizedWork.objects.get(id=form.instance.pk)
         digwork.index(params={"commitWithin": 10000})
 
+    def bulk_add_collection(self, request, queryset):
+        '''
+        Bulk add a queryset of :class:ppa.archive.DigitizedWork to
+        a :class:ppa.archive.Collection.
+        '''
+        # Uses POST from admin rather than a database query to get the pks
```
rlskoeser: I see that in the examples, although I don't see it as a suggested best practice anywhere (so a link could be useful, if there's a good one to provide).

However, I did some testing and it looks like using the selected value rather than the queryset means you don't get extra items beyond the current page if the user clicks the 'select all' option.  I think supporting that would be valuable, because if admins can search on something that will find a large number of items that belong in a collection, they can add them all at once.

ppa/archive/templates/admin/archive/digitizedwork/bulk_add_collections.html
```diff
@@ -0,0 +1,76 @@
...
+            <li><a href="{% url 'admin:index' %}">{% trans "Home" %}</a></li>
+            <li><a href="{% url 'admin:app_list' app_label='archive' %}">Archive</a></li>
+            <li><a href="{% url 'admin:archive_digitizedwork_changelist'%}">Digitized Works</a></li>
+            <li>Bulk Add to Collections</li>
+        </ul>
+    {% endif %}
+{% endblock %}
+
+{% block content_title %}
+    <h1>Bulk Add to Collections</h1>
```
rlskoeser: After using this a couple times, I think "bulk" should be dropped here.  It's doesn't add anything, and is potentially confusing.  Maybe relabel to be more explicit: "Add digitized works to collections"?

I think it's fine to keep the "bulk" terminology internally within the code, but suggest dropping it anywhere else that is public facing or API-ish (like the url names).

rlskoeser:
@bwhicks after thinking about this more, I think you should test and try to make this handle adding all ~5000 items.  Even though that would be a silly use case for adding content to collections, it would give us good confidence that the logic will scale up (and if it doesn't and is difficult to scale up, then let's find out and document the threshold where it will fail, and handle appropriately in the interface).

ppa/archive/tests/test_views.py
```diff
@@ -273,3 +272,112 @@ def test_template(self):
...
+        self.assertContains(response,
+            '<h1>Bulk Add to Collections</h1>', html=True)
+        # check html=False so we can look for just the opening tag of the form
+        # block (html expects the all the content between the closing tag too!)
+        self.assertContains(
+            response,
+            '<form method="post"',
+        )
+        # values are passed to hidden form field correctly
+        print(response.render().content)
```
rlskoeser: debug stray print statement left in

bwhicks:
@rlskoeser I've refactored with an approach that uses Django's session framework to store the `DigitizedWork` ids when passing them from the admin to the custom view. It also successfully handled the entire set of 5,000 works (with a few seconds delay). Now I'll turn to the naming and spot issues, but I think this will fit larger use cases than will likely be done with the bulk add.

ppa/archive/tests/test_admin.py
```diff
@@ -16,13 +18,13 @@
...
 
-@override_settings(SOLR_CONNECTIONS=TEST_SOLR_CONNECTIONS)
-@patch('ppa.archive.solr.get_solr_connection')
 class TestDigitizedWorkAdmin(TestCase):
 
     def setUp(self):
         self.factory = RequestFactory()
 
+    @override_settings(SOLR_CONNECTIONS=TEST_SOLR_CONNECTIONS)
+    @patch('ppa.archive.solr.get_solr_connection')
```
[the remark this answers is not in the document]
bwhicks: I was using a Django form in case the admin method did anything behind the scenes I thought I might miss with the form, but realized after you pointed it out that a simple `Mock()` would do just fine to supply the required `form` parameter to the admin action. Revised accordingly.

ppa/archive/tests/test_admin.py
bwhicks: Given the new framework, I think there likely are not. I handled the error of no pks or bad pks (somehow from a previously used url), but the latter is broadly fixed using the session framework (much harder for a user to set them) and the former I handle in the add view (which displays a message if pks haven't been set).

I will add a test for an empty queryset, though. I'm not sure it would pose an issue, but I see no harm in making sure it behaves as expected.

ppa/archive/views.py
bwhicks: I've attempted to clarify this with a new comment while refactoring.

ppa/archive/views.py
bwhicks: I did indeed miss those. I'll double check building the sphinx doc before finishing work on revisions, but I believe I edited most of the instances of this while refactoring.

ppa/archive/views.py
bwhicks: I revised to do just this. Numbers and collections added to are now noted.

ppa/archive/admin.py
bwhicks: I think in our use case, as you say, the queryset is the preferable way to go. It seemed better to avoid the database ping, but given the need to select results beyond 100 (or whatever upper display limit we configure for the admin) and the fact that the admin action misses them currently, that falls behind the need to do something useful.

ppa/archive/urls.py
bwhicks: Do you have any suggestions for this url terminology or the view name, @rlskoeser, having had some time to mull it?  If not, I'll think and take another stab at it

ppa/archive/admin.py
```diff
@@ -33,16 +33,15 @@ def save_related(self, request, form, formsets, change):
...
         # Uses POST from admin rather than a database query to get the pks
         # per the suggested practices in Django documentation
-        selected = request.POST.getlist(admin.ACTION_CHECKBOX_NAME)
-
-        return HttpResponseRedirect(
-            '%s?ids=%s' % (reverse('archive:bulk-add'), ','.join(selected))
-        )
+        selected = ','.join(str(val) for val in
+                            queryset.order_by('id').values_list('id', flat=True))
+        request.session['selected_works'] = selected
```
rlskoeser: Couldn't you just pass the list and let django serialize it as json for you?

Also, a more specific session key might be a good thing here. Maybe  `collection-bulkadd-ids` or something along those lines?

bwhicks: I was considering a more specific key, so I like that suggestion. 

My thought was to keep it simple with a comma delimited list, but in principle I don't see why that wouldn't work. I doubt the parsing difference would be significant at all if you think a list would be more clear.

rlskoeser: I think it will simplify the code, which will make it more readable. Agreed, probably similar performance.

ppa/archive/urls.py
rlskoeser: What do you think of `add-to-collection` for url name, and `AddToCollection` for the view?

bwhicks: I like it. I'll change the url and view accordingly. Thank you!

ppa/archive/templates/admin/archive/digitizedwork/bulk_add_collections.html
bwhicks: I've revised this for clarity/public-facing terminology.

ppa/archive/templates/archive/add_to_collection.html
```diff
@@ -0,0 +1,71 @@
...
+            <li><a href="{% url 'admin:index' %}">{% trans "Home" %}</a></li>
+            <li><a href="{% url 'admin:app_list' app_label='archive' %}">Archive</a></li>
+            <li><a href="{% url 'admin:archive_digitizedwork_changelist'%}">Digitized Works</a></li>
+            <li>Add DigitizedWorks to Collections</li>
+        </ul>
+    {% endif %}
+{% endblock %}
+
+{% block content_title %}
+    <h1>Add DigitizedWorks to Collections</h1>
```
rlskoeser: minor quibble: `DigitizedWorks` is the name of the model; it would be better to use the verbose/human readable version of that here and elsewhere on user-facing pages.

ppa/archive/views.py
```diff
@@ -166,3 +171,72 @@ def render_to_csv(self, data):
 
     def get(self, *args, **kwargs):
         return self.render_to_csv(self.get_data())
+
+
+class AddToCollection(ListView, FormMixin, ProcessFormView):
+    '''
+    View to bulk add a queryset of :class:`ppa.archive.models.DigitizedWork`
+    to a set of :class:`ppa.archive.models.Collection instances`.
+
```
rlskoeser: Should document here that it expects a list of DigitizedWork ids to be set in the session.

ppa/archive/views.py
```diff
@@ -166,3 +171,72 @@ def render_to_csv(self, data):
...
+    success_url = reverse_lazy('admin:archive_digitizedwork_changelist')
+
+    def get_queryset(self, *args, **kwargs):
+        '''Return a queryset filtered by id, or empty list if no ids'''
+        # get ids from session if there are any
+        ids = self.request.session.get('collection-add-ids', [])
+        # if somehow a problematic non-pk is pushed, will be ignored in filter
+        digworks = DigitizedWork.objects.filter(id__in=ids if ids else [])
+        # revise the stored list in session
+        self.request.session.ids = list(digworks)
```
rlskoeser: I'm not quite following the view logic for why you have to put the ids back in the session.  Is this because you need the list on GET for display and then on POST for processing?  Maybe clarify this in the comments a bit?

ppa/archive/views.py
```diff
@@ -166,3 +171,72 @@ def render_to_csv(self, data):
...
+        form = AddToCollectionForm(request.POST)
+        ids = []
+        if 'collection-add-ids' in request.session:
+            ids = request.session['collection-add-ids']
+        if form.is_valid() and ids:
+            data = form.cleaned_data
+            # clear the session variable to prevent repeat submission
+            del request.session['collection-add-ids']
+            # get digitzed works from validated form
+            digitized_works = DigitizedWork.objects.filter(id__in=ids)
```
rlskoeser: Can you not use `get_queryset` for this?  In fact, I think it may have already run at this point and be stored on your view class somewhere.

PR #28, opened 2018-04-23
Use add for collection bulk_add rather than set

bwhicks:
Minor additions to collection management and bug fix

### DIFF
--- a/ppa/archive/admin.py
+++ b/ppa/archive/admin.py
@@ -9,7 +9,7 @@ class DigitizedWorkAdmin(admin.ModelAdmin):
     list_display = ('title', 'hathitrust_link', 'author', 'list_collections',
         'enumcron', 'pub_place', 'publisher', 'pub_date', 'page_count',
         'added', 'updated')
-    fields = ('hathitrust_link', 'source_url', 'title', 'enumcron', 'author',
+    fields = ('hathitrust_link', 'title', 'enumcron', 'author',
         'pub_place', 'publisher', 'pub_date', 'collections', 'added',
         'updated')
     readonly_fields = ('hathitrust_link', 'page_count',
--- a/ppa/archive/admin.py
+++ b/ppa/archive/admin.py
@@ -6,8 +6,9 @@ from ppa.archive.models import DigitizedWork, Collection
 
 
 class DigitizedWorkAdmin(admin.ModelAdmin):
-    list_display = ('source_id', 'title', 'author', 'enumcron', 'pub_place',
-        'publisher', 'pub_date', 'page_count', 'added', 'updated')
+    list_display = ('source_id', 'title', 'author', 'list_collections',
+        'enumcron', 'pub_place', 'publisher', 'pub_date', 'page_count',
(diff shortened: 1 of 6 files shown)


ppa/archive/admin.py
```diff
@@ -17,6 +18,14 @@ class DigitizedWorkAdmin(admin.ModelAdmin):
     list_filter = ['collections']
     actions = ['bulk_add_collection']
 
+    def list_collections(self, obj):
```
rlskoeser: I might have put the list property on the model rather than here, but I don't know that we would use it anywhere else.  Doesn't seem like any point in trying to share logic with the collection list we use for index data.

ppa/archive/views.py
```diff
@@ -221,7 +221,9 @@ def post(self, request, *args, **kwargs):
             digitized_works = self.get_queryset()
             del request.session['collection-add-ids']
             for collection in data['collections']:
-                collection.digitizedwork_set.set(digitized_works)
+                # add rather than set to ensure add does not replace
+                # previous digitized works in set.
+                collection.digitizedwork_set.add(*digitized_works)
```
rlskoeser: Thanks for the detailed comment

rlskoeser:
Looks good so far.

ppa/archive/admin.py
bwhicks: @rlskoeser That was my thought. I initially had it on the model and then decided that this was a once off use, unless you can think of reasons otherwise.

ppa/archive/admin.py
```diff
@@ -6,9 +6,13 @@
 
 
 class DigitizedWorkAdmin(admin.ModelAdmin):
-    list_display = ('source_id', 'title', 'author', 'enumcron', 'pub_place',
-        'publisher', 'pub_date', 'page_count', 'added', 'updated')
-    readonly_fields = ('source_id', 'source_url', 'page_count',
+    list_display = ('title', 'hathitrust_link', 'author', 'list_collections',
```
rlskoeser: Minor quibble: I've been trying to name things generically (even though we're only working with hathi materials for now) so it will be easier to add other content when we get to that point.  I don't feel strongly about holding this up for a rename though.

ppa/archive/admin.py
```diff
@@ -17,6 +21,23 @@ class DigitizedWorkAdmin(admin.ModelAdmin):
...
 
+    def list_collections(self, obj):
+        '''Return a list of :class:ppa.archive.models.Collection object names
+        as a comma separated list to populate a change_list column.
+        '''
+        return ', '.join([coll.name for coll in
+                          obj.collections.all().order_by('name')])
+    list_collections.short_description = 'Collections'
+
+    def hathitrust_link(self, obj):
```
rlskoeser: This looks great

ppa/archive/admin.py
bwhicks: @rlskoeser I can make the change quickly. Do you have thoughts on the name, i.e. `source_link`?

PR #38, opened 2018-05-07
Feature/sort results

bwhicks:


ppa/archive/forms.py
```diff
@@ -25,9 +25,31 @@ def valid_value(self, value):
 
 class SearchForm(forms.Form):
     '''Simple search form for digitized works.'''
+
+    SORT_CHOICES = [
+        ('relevance', 'relevance'),
+        ('pub_date_asc', 'chronology (earliest to latest)'),
+        ('pub_date_desc', 'chronology (latest to earliest)'),
```
thatbudakguy: these have different human-friendly names (e.g. "Year Oldest-Newest") in the design spec. I'm already editing this stuff for a frontend PR so I can just change it there, though.

bwhicks: @thatbudakguy No reason for me not to nab it now. It'll save you some copy-paste later. Thanks for the spot.

thatbudakguy: speaking of copy-paste, I already had this snippet in there:
```
    SORT_TYPES = {
        'relevance': 'Relevance',
        'year_oldest': 'Year Oldest-Newest',
        'year': 'Year Newest-Oldest',
        'title': 'Title A-Z',
        'title_z': 'Title Z-A',
        'author': 'Author A-Z',
        'author_z': 'Author Z-A'
    }
```
maybe saves you 30sec because the keys are wrong, but whatever

thatbudakguy: funny that we both filled in opposite sides of a virtually identical dict without knowing...

ppa/archive/views.py
```diff
@@ -64,17 +65,35 @@ def get_queryset(self, **kwargs):
             # to support exact phrase searches
             solr_q = 'text:(%s) OR {!join from=srcid to=id v=$join_query}' % query
             # sort by relevance, return score for display
-            self.sort = 'relevance'
+            self.sort = 'Relevance'
```
rlskoeser: After testing it, I think you need to move the default sort logic into the form class because otherwise on the display the current sort option isn't selected.

ppa/archive/views.py
```diff
@@ -64,17 +65,35 @@ def get_queryset(self, **kwargs):
...
         else:
             # no search term - find everything
             solr_q = "*:*"
             # for now, use title for default sort
-            self.sort = 'title'
+            self.sort = 'Title A-Z'
             solr_sort = 'title_exact asc'
             fields = '*'
 
+        if sort:
```
rlskoeser: Would it make sense to move this into the form?

rlskoeser: Could even be a method like `get_solr_sort_field`

ppa/archive/forms.py
```diff
@@ -25,9 +25,31 @@ def valid_value(self, value):
...
+        ('author_asc', 'Author A-Z'),
+        ('author_desc', 'Author Z-A'),
+    ]
+
+    def __init__(self, *args, **kwargs):
+        '''
+        Set choices dynamically based on form kwargs and presence of keywords.
+        '''
+        super(SearchForm, self).__init__(*args, **kwargs)
+        if not args or 'query' not in args[0] or not args[0]['query']:
```
rlskoeser: I like this solution, not what I had thought of.  Does it impact the design at all?  I had imagined relevance being listed but greyed out, but now that I think about it I don't know how that would work.

ppa/archive/forms.py
```diff
@@ -25,9 +25,31 @@ def valid_value(self, value):
...
+            # if there aren't keywords to search for, this will remove
+            # relevance from the form choices
+            self.fields['sort'].choices = self.fields['sort'].choices[1:]
```
rlskoeser: I tested manually submitting the form with sort set to relevance and no keyword and got the generic error message:
> Select a valid choice. relevance is not one of the available choices. 

If there's an easy way to do it, would be nice to customize the error message to add text that relevance is only available when search terms are used - but if it's not easy (and I can't think of an easy way off hand!), then we'll skip it for now.

bwhicks: I bet I can do something in `form.clean()` on this. Let me think about a sane way to handle that scenario.

ppa/archive/forms.py
bwhicks: @thatbudakguy @xinyil A quick question on this for when I pick this back up and not wholly clear from the design spec (as far as I can see): when relevance isn't a valid option (i.e. no keyword search made), should it be in the form choices but disabled in some way or just not there at all (how I have it currently written)?

thatbudakguy: I kind of like it being disabled, personally. moves less things around and easier to indicate in the UI.

thatbudakguy: Kind of a bummer that we have to do that, but I think it'd work. As long as it supports disabling I don't think the widget type matters much because I have to iterate through this thing to make it display correctly (it's not going to render as radio buttons or a select anyway).

ppa/archive/forms.py
```diff
@@ -81,6 +79,32 @@ def __init__(self, *args, **kwargs):
         'collections_exact': 'collections'
     }
 
+    def get_solr_sort_field(self, query, sort):
```
rlskoeser: it looks like `query` variable unused here?

ppa/archive/views.py
```diff
@@ -32,14 +32,30 @@ class DigitizedWorkListView(ListView):
...
 
     def get_queryset(self, **kwargs):
-        self.form = SearchForm(self.request.GET)
-        query = join_q = collections = None
+        # Check for a nonsensical search with 'relevance' and no query string
+        # and amend GET dictionary, first removing any blank query strings
+        # so we can just ask if the key exists
+        # testing for empty string specifically to avoid removing integers
+        # for any future query strings
+        GET = {k: v for k, v in self.request.GET.dict().items() if v != ''}
```
rlskoeser: Pretty sure this is going to cause problems for fields that can be multiple, because I struggled with that in derrida.  See https://github.com/Princeton-CDH/derrida-django/blob/master/derrida/books/views.py#L119-L129

Can we borrow any of the derrida default logic to simplify this?

ppa/archive/forms.py
```diff
@@ -36,6 +79,35 @@ class SearchForm(forms.Form):
...
+            'title_desc': 'title_exact desc',
+            'author_asc': 'author_exact asc',
+            'author_desc': 'author_exact desc',
+        }
+        template_mapping = dict(self.SORT_CHOICES)
+        fields = '*'
+
+        if sort == 'relevance':
+            # relevance requires we include score field
+            fields = '*,score'
```
rlskoeser: I think the fields logic needs to be in the view, not the form (esp. since I've been thinking I'd like to pare down the fields to just the ones we actually need, instead of returning everything).  If it's any easier, we could just include score whenever there's a keyword query (not sure if there's a cost associated with calculating it?).

ppa/archive/tests/test_views.py
```diff
@@ -178,10 +178,58 @@ def test_digitizedwork_listview(self):
         response = self.client.get(url, {'query': '"incomplete phrase'})
         self.assertContains(response, 'Unable to parse search query')
 
+        # add a sort term - pub date
+        response = self.client.get(url, {'query': '', 'sort': 'pub_date_asc'})
+        # explicitly sort by pub_date manually
+        sorted_object_list = sorted(response.context['object_list'],
```
rlskoeser: Would it make more sense to get a sorted list from the db and compare?

ppa/archive/forms.py
bwhicks: I had pulled it to the form because it was dependent on the field set, but I think if we use the second option (and probably minimal cost) we can simplify the code significantly.

Issue #31, opened 2018-05-02
As a user viewing keyword search results, I want to see a few text snippets from the full text of a work so that I can get an idea how my search terms are used in the work.


thatbudakguy:
## Notes for testing
- you should not see page images or text highlighting if you don't enter search terms
- you should see one or two page images with text highlighting when you enter search terms; most relevant pages are displayed first; for now, matching terms will be italicized (styles will be added later based on Xinyi's design)
- you should see a link "view all N pages" only if there are more than two matches; link goes to the item detail view for now (actual functionality will be implemented and tested under #32)

meagan1019:
Searched keyword "saintsbury" with book title "poetry" and returned 575 works. (Note: book title and author search seem disabled currently. Searching keyword "saintsbury" alone also returned 575 works. See my slack comment.) Keyword returned in italics, first two images were displayed, and a link for additional page images/results per work was populated (linked to bib entry of work currently). Did the same for searching "saintsbury AND meter" and it worked perfectly!

PR #43, opened 2018-05-08
Feature/keyword search snippets

rlskoeser:
Feature #31 

- adds convenience methods to local solr query object to provide access to expanded & highlighting results
- adds custom templatetag for getting a dict item from a variable key, to simplify templates for expanded results and highlights
- adds view logic to query solr for pages with highlighted terms; currently requesting 3 snippets, but we can revisit
- basic template display of page images and text snippets
- does not include page label included in Zeplin mockup because we don't have access to that in the current import/index

### DIFF
--- a/ppa/archive/templatetags/ppa_tags.py
+++ b/ppa/archive/templatetags/ppa_tags.py
@@ -1,0 +1,10 @@
+from django.template.defaulttags import register
+
+@register.filter
+def dict_item(dictionary, key):
+    ''''Template filter to allow accessing dictionary value by variable key.
+    Example use::
+
+        {{ mydict|dict_item:keyvar }}
+    '''
+    return dictionary.get(key, None)
--- a/ppa/archive/views.py
+++ b/ppa/archive/views.py
@@ -31,11 +31,14 @@ class DigitizedWorkListView(ListView):
 
     paginate_by = 50
 
+    # keyword query; assume no search terms unless set
+    query = None
+
     def get_queryset(self, **kwargs):
         self.form = SearchForm(self.request.GET)
-        query = join_q = collections = None
+        join_q = collections = None
         if self.form.is_valid():
-            query = self.form.cleaned_data.get("query", "")
+            self.query = self.form.cleaned_data.get("query", "")
             # NOTE: This allows us to get the name of collections for
             # collections_exact and set collections to a list of collection names
             collections = self.form.cleaned_data.get("collections", None)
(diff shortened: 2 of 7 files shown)


rlskoeser:
A couple things occurred to me about this code: 
1. image url is pretty hathi specific; any ideas how to make it easier to generalize later?
2. I don't think I wrote any tests for the template logic, and probably should before this is complete

ppa/archive/templatetags/ppa_tags.py
```diff
@@ -0,0 +1,33 @@
...
+    Example use::
+
+        {{ mydict|dict_item:keyvar }}
+    '''
+    return dictionary.get(key, None)
+
+
+@register.simple_tag(takes_context=True)
+def querystring_replace(context, **kwargs):
+    '''Template tag to simplify retaining querystring parameters
```
thatbudakguy: this is pretty cool!

ppa/archive/views.py
```diff
@@ -91,22 +94,68 @@ def get_queryset(self, **kwargs):
...
+            'expand.rows': 2,   # number of items in the collapsed group, i.e pages to display
             'join_query': join_q,
-            'coll_query': coll_query
-            # 'rows': 50  # override solr default of 10 results; display 50 at a time for now
+            'coll_query': coll_query,
+            # enable highlighting
+            'hl': True,
+            'hl.fl': 'content',
+            # override solr default of 10 results; display 50 at a time for now
+            # 'rows': 50
```
thatbudakguy: are we not defaulting to 50 anymore?

ppa/archive/views.py
```diff
@@ -91,22 +94,68 @@ def get_queryset(self, **kwargs):
             'facet.field': [field for field in self.form.facet_fields],
             # default expand sort is score desc
             'expand': 'true',
-            'expand.rows': 10,   # number of items in the collapsed group, i.e pages to display
+            'expand.rows': 2,   # number of items in the collapsed group, i.e pages to display
             'join_query': join_q,
-            'coll_query': coll_query
-            # 'rows': 50  # override solr default of 10 results; display 50 at a time for now
+            'coll_query': coll_query,
+            # enable highlighting
```
thatbudakguy: do we need highlighting enabled here even though we're doing it using another query?

ppa/archive/views.py
rlskoeser: whoops, good catch.

ppa/archive/views.py
rlskoeser: Nope! Thanks for catching that

Issue #30, opened 2018-05-02
As a user, I want to filter search results by publication year or range of years so that I focus on works from a particular time period.


thatbudakguy:
## Notes for testing
- placeholder text should display max/min values based on what's in the database
- should be able to specify just a min (everything after date), just a max (everything before), or both
- use same date in both to get items from a single year
- entering a min greater than max should give an error message
- entering a date outside the max/min in the db should give an error message

meagan1019:
Searched "(iamb OR trochee) AND count" returned 200 works. 1559 and 1922 were pre-populated, which are the correct dates! Filtered with start date of 1891 (no end date), works dropped to 112. Filtered with end date of 1820, works dropped to 5. Search 1820-1820, returned the single work published that year. Received error message when I tried to filter outside of parameters (entered 1500 and 1930). However, the thumbnails and text results with italicized search terms did not display when filtering by year (in any iteration I tested). Once I entered anything in the publication year fields (min and/or max) the results no longer displayed the actual results, just the work bibliography.

rlskoeser:
I still didn't fix this properly.  Back in dev queue and not ready for testing.

meagan1019:
searched "iamb AND trochee" from 1900 to 1920 (and with a min only and a max only; and within a single year). Three snippets work great! Tried entering a reverse range (1900 to 1780) and also received the appropriate error message.

PR #52, opened 2018-05-09
Feature/date range filter

rlskoeser:
Publication date range filter for #30


ppa/archive/forms.py
```diff
@@ -53,6 +114,45 @@ def set_choices_from_facets(self, facets):
...
+    def pub_date_minmax(self):
+        '''Get minimum and maximum values for
+        :class:`~ppa.archive.models.DigitizedWork` publication dates
+        in the database.  Used to set placeholder values for the form
+        input and to generate the Solr facet range query.
+        Value is cached to avoid repeatedly calculating it.
+
+        :returns: tuple of min, max
+        '''
+        # maxmin = cache.get(self.PUBDATE_CACHE_KEY)
```
bwhicks: Am I looking at this wrongly or would this line being commented out result in `pub_date_minmax` never pulling in the value from cache?

rlskoeser: @bwhicks whoa, thanks for catching that.  I must have commented it out when I was testing some error handling.

ppa/archive/views.py
```diff
@@ -81,7 +87,47 @@ def get_queryset(self, **kwargs):
...
+
+            # get minimum and maximum pub date values from the db
+            pubmin, pubmax = self.form.pub_date_minmax()
+
+            start = int(start) if start else pubmin
+            end = int(end) if end else pubmax
+
+            # fallback logic for when start and end are not set
+            # (should only happen when no content is in the database)
+            if not start or end:
```
thatbudakguy: how can we get here after lines 105-106? if start/end are `None` won't they be set above?

thatbudakguy: is this supposed to be `if not pubmin or pubmax`, etc?

rlskoeser: Good point.  Just simplified to 
```
start = int(start) if start else pubmin or 0
end = int(end) if end else pubmax or 1922
```

Issue #32, opened 2018-05-02
As a user viewing an individual item from a keyword search, I want to see page image thumbnails and text snippets that match my search terms so I can see how many and what kind of pages match my search terms.


thatbudakguy:

meagan1019:
<img width="1049" alt="screen shot 2018-05-15 at 3 28 41 pm" src="https://user-images.githubusercontent.com/4306343/40079150-e45799ac-5854-11e8-9bac-88b28f02331c.png">
search ss

meagan1019:
<img width="1088" alt="screen shot 2018-05-15 at 3 28 55 pm" src="https://user-images.githubusercontent.com/4306343/40079196-03a29730-5855-11e8-94cd-ea8055730790.png">
first return with only 1 term highlighted/displayed as thumbnail

meagan1019:
<img width="1102" alt="screen shot 2018-05-15 at 3 29 25 pm" src="https://user-images.githubusercontent.com/4306343/40079212-1a5e8290-5855-11e8-840f-8271b632cb91.png">
second return with 2/3 terms highlighted/displayed as thumbnail

meagan1019:
<img width="1095" alt="screen shot 2018-05-15 at 3 29 54 pm" src="https://user-images.githubusercontent.com/4306343/40079233-2b25bd5a-5855-11e8-88a8-4099c29373d2.png">
fourth return (the only) with all three terms displayed

rlskoeser:
@meagan1019 I see what you mean now, thanks for documenting!  Each of these pages have all of your search terms, but we're currently only displaying the first highlighted snippet (which may or may not contain all, apparently!).  We had it set to 3 snippets for a while, but that doesn't work too well in the current design for some of the results.  Do you want me to set it back to 3 so you can test it that way?

meagan1019:
earlier test case fixed. searched "raven and poe" since the previous search didn't return multi-page results from the same work. clicked on "view all 8 pages" for mdp.39015048893823 (result 3 by relevance). Was clearly able to see all 8 pages with my search terms italicized. The (AMAZING) search bar at the top was there, and I searched the term "triple," which returned two new pages with terms highlighted. I am in love with this feature!

meagan1019:
Before I close, the testing asks if there is "a paginated list of matching pages" when you click on the "view all N pages" from search result page. I'm struggling to see where the page number is displayed outside of the image thumbnail itself. Or is that where the pagination comes in? 
[...]

PR #60, opened 2018-05-11
Feature/thumbnail snippets detail

bwhicks:
This pull request provides basic functionality for #32 (keyword search, thumbnails and snippets on detail view)


ppa/archive/views.py
```diff
@@ -247,7 +253,56 @@ class DigitizedWorkDetailView(DetailView):
...
+            solr_opts = {
+                'q': solr_q,
+                'sort': 'order asc',
+                'fl': '*',
+                'hl': True,
+                'hl.fl': 'content',
+                # No way to set this to actual number in one search
+                # therefore using an arbitrarily high but not system breaking
+                # number.
+                'hl.snippets': 10000,
```
rlskoeser: This number is how many snippets to return for each item that matches, so it really doesn't need to be big.  Probably only 2-3 ?

ppa/archive/templates/archive/snippets/thumbnail_snippet.html
```diff
@@ -0,0 +1,16 @@
+<div style="float:left; width: 50%">
```
rlskoeser: Please add a comment block at the top here explaining briefly what this is for and what variables it expects.

ppa/archive/views.py
```diff
@@ -247,7 +253,56 @@ class DigitizedWorkDetailView(DetailView):
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        # pull in the query if it exists to use
+        query = self.request.GET.get('query', '')
+
+        solr_pageq = None
+        if query:
+            context['query'] = query
+            digwork = context['object']
+            solr_q = 'text:(%s) AND srcid:("%s") AND item_type:(page)' \
```
rlskoeser: I recommend using srcid and item type query as filter queries. No need for `()` around page since it's a single term with no weird punctuation or anything like that.

ppa/archive/views.py
```diff
@@ -247,7 +253,56 @@ class DigitizedWorkDetailView(DetailView):
...
+                % (query, digwork.source_id)
+            solr_opts = {
+                'q': solr_q,
+                'sort': 'order asc',
+                'fl': '*',
```
rlskoeser: Let's limit the fields to what we actually need for display (maybe just id right now?); in particular I'd like to not include `content`, since that makes the result so much smaller.

ppa/archive/templates/archive/snippets/thumbnail_snippet.html
```diff
@@ -0,0 +1,16 @@
...
+        srcset="{% page_image_url id order 360 %} 2x"
+        style="float:left"/>
+          {# the listview use of this snippet passes in a a list of highlights#}
+          {% if highlights %}
+            {% for snippet in highlights.content %}
+              <p style="white-space: pre-wrap">{{ snippet|safe }}</p>
+            {% endfor %}
+          {% endif %}
+          {# the detail view version is only concerning itself with pages so has just one item #}
+          {% if item %}
```
rlskoeser: I don't think we ever want to display the full text content of the page

ppa/archive/templates/archive/digitizedwork_detail.html
```diff
@@ -105,6 +105,45 @@ <h2 class="header">{{ object.title }}</h2>
...
 </div>
+
+{% if error %}
+  {# if something went wrong (e.g., bad query), display error message and nothing else #}
+  <p>{{ error }}</p>
+{% else %}
+  {% if page_highlights %}
+
+    {% for item in page_highlights %}
+        {% include 'archive/snippets/thumbnail_snippet.html' with id=item.srcid order=item.order item=item%}
```
rlskoeser: I think this isn't actually setting the highlight information properly for display in the snippet template.  Or, that is - it's not setting it in the context the same way it's being set for the archive view (which makes sense because in that case we're running two separate queries that we have to piece back together).

ppa/archive/views.py
```diff
@@ -247,7 +253,56 @@ class DigitizedWorkDetailView(DetailView):
...
+
+            try:
+                # configure paginator and set in context
+                solr_pageq = PagedSolrQuery(solr_opts)
+                paginator = Paginator(solr_pageq, per_page=self.paginate_by)
+                page = self.request.GET.get('page', 1)
+                context['page_obj'] = paginator.page(page)
+                # get highlights and set in context
+                highlights = solr_pageq.get_results() if solr_pageq else []
+                context['page_highlights'] = highlights
```
rlskoeser: Ah, I see.  You also need to call `get_highlighting` to get the highlight snippet results returned by Solr.  So here, `highlights` is actually your matching pages and you're not passing in the highlight information to the template for display.

PR #72, opened 2018-05-16
Simplify compressor pipelines; allow ES6

thatbudakguy:
- moves js dependencies to `package.json` file and adds lockfile
- updates README to specify installation step
- updates travis to use new installation step
- uses `django-compressor-toolkit` pipeline for scss (identical to our old one - allows `@import`s via `/static` & uses autoprefixer)
- uses `django-compressor-toolkit` pipeline for js (allows es6 w/ modules via browserify & babel)

note that to trigger compressor js files need to have the `type="module"` attribute, e.g.:
```py
{% compress js %}
    <script src="{% static 'js/index.js' %}" type="module"></script>
{% block js %}
```


package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "ppa-django",
+  "description": "Django web application for the Princeton Prosody Archive",
+  "repository": "https://github.com/Princeton-CDH/ppa-django.git",
+  "author": "The Center for Digital Humanities at Princeton <cdh@princeton.edu>",
```
rlskoeser: cdh@princeton isn't actually one of our email addresses - you could use cdh-info, but may cdhdevteam might make more sense here

PR #75, opened 2018-05-17
Feature/histogram

thatbudakguy:
- adds basic histogram using canvas API
- component receives data from hidden dl/dt elements that could be used to improve accessibility of the visualization?
- cleans up some styles in the form JS in prep for refactor

note @rlskoeser I probably should've used these conventions sooner and more explicitly - if you're okay with them:
- `myVar` is any variable
- `$myVar` is some `HTMLElement`, e.g. `document.getElementsById('myVar')` or `$('#myVar').get()`
- `$$myVar` is the jQuery object creating from selecting some element, e.g. `$('#myVar')`

for latter two I usually use `const` to indicate they're an unambiguous reference that gets evaluated once (on page load) only


rlskoeser:
Looks reasonable and seems to be working pretty well.  I like the idea of the hidden structured values for screen readers and such.

Can we set a minimum width for the bar? i.e., if we only have one or two data points, it shouldn't take the full width.

How hard would it be to add hover text with year(s) and count?

Naming conventions sound reasonable to me.  Should we document them somewhere?  (maybe team documentation rather than project specific)

thatbudakguy:
setting a max width for the bar seems tricky because if we're still going to display the min/max dates at the edges of the visualization, they won't match up with the actual locations of bars when there are few bars. does that make sense?

i think the visualization works fine for any decent size ranges - for very small ranges, people probably aren't looking to it anyway since the result set should be a fraction of a page. even if it looks a little silly to only have two bars or to have the visualization be "full", I think visual consistency is better than trying to adjust it to handle small ranges.

I haven't tried to make the bars interactive yet, but I can't imagine it's that difficult. I think that functionality can come in the next iteration of the search form.

I think maybe I can keep a running note of JS conventions I'm using for this project and convert it into a doc that can live in the drive or somewhere so we can use it for future projects. we could also use a linter with some settings; I like eslint (but that might be overkill).

PR #81, opened 2018-05-18
Break out import and index logic into two scripts, improve index speed

rlskoeser:
- hathi_import script now *only* imports, does not index
- new index script can index everything or only works, only pages
- new index script is much faster
- hathi pairtree logic has been refactored to be more reusable, should be better tested
- multiprocessing out of scope for now (and maybe unnecessary with speed improvements)


ppa/archive/management/commands/index.py
```diff
@@ -0,0 +1,154 @@
...
+    hathi_pairtree = {}
+    stats = None
+
+    options = {}
+    #: normal verbosity level
+    v_normal = 1
+    verbosity = v_normal
+
+    #: solr params for index call; currently set to commit within 5 seconds
+    solr_index_opts = {"commitWithin": 5000}
```
thatbudakguy: random thought: you could make this a CLI argument too? not sure how much it matters, but if differences in RAM across machines actually do affect it substantially, this would allow you to play around until you got a workable solution

ppa/archive/management/commands/index.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'source_ids', nargs='*',
+            help='Optional list of specific works to index by source (HathiTrust) id.')
+        parser.add_argument(
+            '-i', '--index', choices=['all', 'works', 'pages'],  default='all',
+            help='Index only works or pages (by default indexes all)')
+        parser.add_argument(
+            '--no-progress', action='store_true',
```
thatbudakguy: 👍  for defaulting to progressbar :)

rlskoeser: @thatbudakguy inspired by your feedback on the import script.

There's no summary output in this one - should there be?

ppa/archive/management/commands/index.py
rlskoeser: Nice idea.  Option should be in seconds probably?  I tried 5 and 3 and didn't see much difference between them for my dev setup.

bwhicks: I like the idea of this being a command line argument, though the good thing is that the current number did work on the default (minimal) JVM heap. So we may just have a good default already.

ppa/archive/management/commands/index.py
```diff
@@ -0,0 +1,154 @@
...
+        # NOTE: currently no good way to catch a connection
+        # error when Solr is not running because we get multiple
+        # connections during handling of exceptions.
+        try:
+            self.solr.index(self.solr_collection, index_data,
+                            params=self.solr_index_opts)
+        except Exception as err:
+        # except (ConnectionError, RequestException) as err:
+            # NOTE: this is still pretty ugly; what part should we return?
+            raise CommandError(err)
```
thatbudakguy: does it have the gnarly java stacktrace? that seems like something we probably don't need (all of). Conversely, the underlying java exception class (e.g. is it an OOMError or not) is probably good to see

bwhicks: I would tend to argue for a full stacktrace, gnarly though it be (Java/Scala, here's looking at you).

ppa/archive/management/commands/index.py
bwhicks: I don't think a summary output is necessary, since we can ping the solr server for the outcome and number of docs it currently has, etc. The progressbar is more meaningful in day to day use imo.

If it's a quick add though, more info is better than less.

ppa/archive/management/commands/index.py
rlskoeser: The problem is that there are multiple exceptions happening in the processing of other connections - this is also why I had such a hard time figuring out the right exception to even catch.  There seem to be some in SolrClient, some in requests, and some in urllib3.

ppa/archive/management/commands/index.py
thatbudakguy: @rlskoeser that sounds right. and definitely agree @bwhicks 5sec can be the default

PR #83, opened 2018-05-21
Feature/install semantic

thatbudakguy:
lots of new files here (semantic UI's source) but not a ton really going on. tagged @bwhicks since it sort of has infrastructure implications. basically:

- semantic gets installed with `npm` along with the rest of the js dependencies; nothing new there. however it does put itself outside of `node_modules` so we can modify its source.
- there's a file `semantic.json` in the root that tells it where to put the source (`sitemedia/semantic/src`) and where to output the compiled stylesheets and js (`sitemedia/semantic/dist`).
- not checking in the `sitemedia/semantic/dist` folder; we just need to run the build step on each deploy to make sure it gets populated...
- ...and this should happen automatically because I made an npm task to do it which will get run during npm's `postinstall` step, which happens once each deploy.
- instead of linking the prebuilt "default" stylesheets and js for semantic, which I was using to prototype, we're now linking to ones we build ourselves instead (changed in `base.html`)

also added some README notes about the build and the npm tasks I added to `package.json`.


thatbudakguy:
@rlskoeser awesome, thanks so much for testing it!

thatbudakguy:
@bwhicks perfect, that was my hope. thanks for checking!

PR #91, opened 2018-05-24
@retry decorator

thatbudakguy:
docblock on the `Retry` class explains how it's supposed to work. I implemented something like this once before, so I was hoping to adapt the code and make it just "smart" enough to distinguish if it was wrapping an admin command's `handle()` and use the provided reference to stdout since I saw our commands do that.

if there's equivalent functionality in `requests` we probably don't need @retry, but if it worked right it could handle non-http related errors too (although idk what "retrying" those is useful for unless you're trying to brute-force a race condition problem).

I started to write a test but realized it probably wasn't the right way to go...I think maybe you could test by faking a slow `HTTPRequest`, but I'm not sure how to emulate a situation where a request fails a few times but finally succeeds.


rlskoeser:
The retry decorator is neat, but I don't think we'd want to retry the whole command.  If you are importing thousands of Hathi items and got an error part way through, wouldn't that start over with reimporting all of them, even the ones that succeeded?

Here's an answer on stack overflow that looks like a neat solution if we want more fine-grained retry control at the request layer: https://stackoverflow.com/a/35636367/9706217

Where were you getting the connection errors? Are you able to duplicate it or was it a transient issue?

thatbudakguy:
@rlskoeser maybe for that case you'd just add the decorator to whatever internal function was being iterated on instead of the entire `handle()` call, so in this case it'd be
```py
@retry
def index(self, index_data):
...
```

thatbudakguy:
I'm not having any issues that require this anymore and I can't really test it, so I'll shelve it.

Issue #65, opened 2018-05-15
As a user, I want my results updated as soon as I change my search options so that I can refine my search without losing my place in the form and have results load more quickly.


thatbudakguy:
## testing notes
- no need to hit <kbd>enter</kbd>!
- the histogram should update when your results change.
- the result count should update when your results change.
- pagination controls should update when your results change.
- clicking "clear selection" on the dates should update the results.
- changing the sort or the collection filters should update the results.
- if you don't have anything entered in keyword, title, or author, you shouldn't be able to sort by relevance (that wouldn't make any sense!)
- if you have relevance selected as a sort and you delete all your queries, you should get switched to title (a-z) sort instead.

jabauer:
On iOS "done" button does not submit the form, have to hit "return" on virtual keypad. On Android "enter" moves to the next form field but does not submit.

xinyil:
submitting when clicking out the form would be good!

thatbudakguy:
ok, so as per #63 I know I need to make the publication date inputs submit when you unfocus them. We probably shouldn't do that for the text inputs though because it would mean you couldn't do e.g. an author and title query at the same time (which works great currently).

If I understand you @xinyil we could make clicking in a place that isn't the form (anywhere else on the page) submit the form, which I think would work but maybe isn't what the user would expect for mobile? I'm curious why mobile currently doesn't show a "submit" button in the native interface - only ways to move back and forth between fields (@rlskoeser and I confirmed this on android too). The form will submit if you get to the last field and hit "next", but only then.

Currently there's actually a hidden submit button, which is why pressing "enter" on the text fields works...just don't know why mobile isn't picking it up.

xinyil:
hitting "done" each form on iOS should submit the form. 

related discussions:
https://stackoverflow.com/questions/5665203/getting-iphone-go-button-to-submit-form?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa
> On a whim, we changed display: none to margin-left: -1000px
> Apparently, Safari is looking for the presence of a SUBMIT button in the form and only if it's not display: none, it will then fire it when you hit the GO button.

thatbudakguy:
@xinyil whoa this is super interesting, thanks for sharing. I'll try using the margin hack to make the button offscreen but still "visible".

thatbudakguy:
I'll update the testing notes as we nail down how we want this to work.

thatbudakguy:
[...]

PR #92, opened 2018-05-29
Feature/submission logic

thatbudakguy:
refs #65 
- search results updated dynamically as user interacts with form
- histogram, result counts, pagination, etc. all update dynamically
- adds template for ajax rendering of results
- refactors search form & histogram classes to make them more general & interoperable
- adds a template filter to enable dumping python objs as json for easier consumption via js


rlskoeser:
Documenting what we discussed: the histogram & reactive form seem like good candidates for unit testing, but not sure we need to do that now. But it might be a good idea to figure out what it looks like for PPA before starting similar javascript work on another project.

rlskoeser:
Finally had a chance to try the code out locally. Generally, seems to be very slick and is working great. The form submits when I add, edit or remove text or change publication years.  Publication year histogram and record counts are updating when the results change. Query string also reflects changes to the form submission, which is great.

- collections and sort results aren't triggering the form to submit on Chrome or Safari, but they do work in Firefox; tried making the checkbox visible in Chrome to see if that made any difference, but it didn't seem to matter.
- relevance sort should be enabled/disabled based on the presence of a keyword search term
- can we revise the query string to only include form fields that have values?

ppa/archive/templates/archive/snippets/results_list.html
```diff
@@ -0,0 +1,23 @@
+{% load static humanize ppa_tags %}
```
rlskoeser: Perhaps a template comment here to indicate this template can be rendered in the full list or displayed via ajax?

ppa/archive/templatetags/ppa_tags.py
```diff
@@ -76,3 +78,10 @@ def solr_highlight(value, autoescape=True):
         mark_safe(part) if EM_TAG_RE.match(part) else esc(part)
         for part in EM_TAG_RE.split(value)
     ]))
+
+
+@register.filter(name='json')
+def json_dumps(data):
```
rlskoeser: Annoying we have to do this, but seems like a reasonable approach.

ppa/archive/tests/test_views.py
```diff
@@ -418,13 +418,37 @@ def test_digitizedwork_listview(self):
         assert not response.context['object_list'].count()
         self.assertContains(response, 'Invalid range')
 
+        # ajax request for search results
+        response = self.client.get(url, HTTP_X_REQUESTED_WITH='XMLHttpRequest')
+        assert response.status_code == 200
+        # should render the results list partial and single result partial
+        self.assertTemplateUsed('archive/snippets/results_list.html')
+        self.assertTemplateUsed('archive/snippest/search_result.html')
```
rlskoeser: Also check `assertTemplateNotUsed` on `archive/list_digitizedworks.html`

ppa/archive/tests/test_views.py
```diff
@@ -418,13 +418,37 @@ def test_digitizedwork_listview(self):
...
+        # shouldn't render the search form
+        self.assertNotContains(response, '<form class="ui form">', html=True)
+        # should have all the results
+        assert len(response.context['object_list']) == len(digitized_works)
```
rlskoeser: add basic sanity check for the data that needs to be present for the ajax form logic to update count and histogram

ppa/archive/tests/test_views.py
```diff
@@ -418,13 +418,37 @@ def test_digitizedwork_listview(self):
...
+        response = self.client.get(url,
+            {'query': 'blood AND bone AND alternate'},
+            HTTP_X_REQUESTED_WITH='XMLHttpRequest')
+        self.assertContains(response, '1 digitized work')
+        self.assertContains(response, wintry.source_id)
+        response = self.client.get(url,
+            {'query': 'blood NOT bone'},
+            HTTP_X_REQUESTED_WITH='XMLHttpRequest')
+        self.assertContains(response, 'No matching works.')
+        response = self.client.get(url,
```
rlskoeser: Not sure how much we gain from testing more than one search query, since it's the same view logic we've already tested pretty thoroughly -- unless there are any variations that impact the ajax behavior or work differently.

ppa/archive/templatetags/ppa_tags.py
rlskoeser: Needs a unit test

ppa/archive/views.py
```diff
@@ -32,6 +32,12 @@ class DigitizedWorkListView(ListView):
     # keyword query; assume no search terms unless set
     query = None
 
+    def get_template_names(self):
+        # allow querying just the results via ajax
```
rlskoeser: I would rephrase this comment to make it more specific, something like: when queried via ajax, return partial html for just the results section without the form

ppa/archive/static/archive/js/search.js
```diff
@@ -1,63 +1,71 @@
...
-        $(event.target).find('input').attr('checked', '')
+    function submitForm(state) {        
+        let url = `?${$.param(state)}` // serialize state using $.param to make querystring
+        window.history.pushState(state, 'PPA Archive Search', url) // update the URL bar
+        let req = fetch(`/archive/${url}`, { // create the submission request
+            headers: { // this header is needed to signal ajax request to django
+                'X-Requested-With': 'XMLHttpRequest'
+            }
+        })
+        req.then(res => res.text()).then(html => { // submit the form and get html back
```
rlskoeser: Nice. Appreciate the comments here

ppa/archive/static/archive/js/search.js
```diff
@@ -1,63 +1,71 @@
...
+            $$paginationTop.html($(html).find('.pagination').html()) // update the top pagination
+            dateHistogram.data = JSON.parse($(html).find('pre').html()) // update the histogram
+            $$resultsCount.html($(html).find('.data .results-count').html()) // update the results count
+            $$results.html(html) // update the results
+            document.dispatchEvent(new Event('ZoteroItemUpdated', { // notify Zotero of changed results
```
rlskoeser: Saw this and remembered to test the Zotero integration - it is updating properly and getting results for current items after the ajax form submission.

thatbudakguy: awesome!!

ppa/archive/tests/test_views.py
thatbudakguy: I think I tried that but it's not a valid assertion?? I couldn't find an idiomatic way to represent that...but I'll take another look

rlskoeser: Does this not work? https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.SimpleTestCase.assertTemplateNotUsed

ppa/archive/static/archive/js/search.js
rlskoeser: @thatbudakguy should we suggest updated documentation for the Zotero docs? Not sure if they have an easy mechanism for that. Here's where I found the old code I shared with you before: https://www.zotero.org/support/dev/exposing_metadata

ppa/archive/tests/test_views.py
thatbudakguy: huh, that definitely looks right. I think either I was trying `assertNotTemplateUsed`, or I did try that but Django was saying that the template _was_ used...I'll try it again.

ppa/archive/static/archive/js/search.js
thatbudakguy: at least one of the methods in their sample has this deprecation message on MDN:
>This feature has been removed from the Web standards. Though some browsers may still support it, it is in the process of being dropped. Avoid using it and update existing code if possible; see the compatibility table at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

so, probably worth letting them know at least? do you know if we could contact someone, or file a PR? don't know where the docs are hosted or if they're public...

rlskoeser:
Collection and sort still not firing a form submission for me in Chrome, but since they're working for @thatbudakguy going to merge this in so we can have more people try it out.

PR #93, opened 2018-06-07
Generalized Solr indexing with signal handling

rlskoeser:
Provides generalized indexing and signal handling code based on the prototype @bwhicks created in the feature/solr-signals branch.

Adds an `Indexable` class with some methods for indexing, removing from index, and indexing chunks of items.  Adds signal handling code for save and delete on the indexable model and a way of identifying dependencies that affect a model's index to provide a generic way of listening to m2m changes and handling related object modification.

My hope is that we could copy and adapt this code in the Winthrop codebase and then have enough information to generalize it.


rlskoeser:
@bwhicks I think this is handling nearly everything that your prototype code handled.  In my testing, I found that I didn't actually need to run the Solr `optimize` call, so I haven't included it here.

README.rst
```diff
@@ -112,7 +112,13 @@ things easier. To run them, first install development requirements::
 
 Run tests using py.test::
 
-    py.test
+    python -m pytest
```
bwhicks: Could I clarify the purpose of this change? It doubtless makes sense but isn't clear to me.

ppa/archive/management/commands/index.py
```diff
@@ -70,8 +69,14 @@ def add_arguments(self, parser):
             'source_ids', nargs='*',
             help='Optional list of specific works to index by source (HathiTrust) id.')
         parser.add_argument(
-            '-i', '--index', choices=['all', 'works', 'pages'],  default='all',
+            '-i', '--index', choices=['all', 'works', 'pages'], default='all',
             help='Index only works or pages (by default indexes all)')
+        parser.add_argument(
```
thatbudakguy: r&d day thought: allow passing any qualified model name as an arg, e.g.
```py 
python manage.py index ppa.archive.works
```
throw an error if the model doesn't implement `Indexable`.

ppa/archive/management/commands/index.py
```diff
@@ -137,18 +131,17 @@ def handle(self, *args, **kwargs):
...
         # error when Solr is not running because we get multiple
         # connections during handling of exceptions.
         try:
-            self.solr.index(self.solr_collection, index_data,
-                            params=self.solr_index_opts)
+            # index in chunks and update progress bar if there is one
+            return Indexable.index_items(index_data, progbar=progbar)
         except Exception as err:
         # except (ConnectionError, RequestException) as err:
             # NOTE: this is still pretty ugly; what part should we return?
```
thatbudakguy: does this still need to be resolved, or is the output OK?

rlskoeser: I think the output is good enough - at least it stops the script instead of repeatedly generating errors.  Maybe something we can keep an eye on as we refine this code.  I'll add some comments to clarify.

README.rst
rlskoeser: @bwhicks pytest can't load the plugin that disconnects the index signal handlers unless it is on the python path.  You can set the environment variable (which is what I'm doing), but for convenience, if you run it this way you get the current directory added to your python path.  I'll add a bit more content to clarify that.

This won't be problem once we refactor/generalize this code into a separate module.

PR #94, opened 2018-06-10
a11y tests & tweaks

thatbudakguy:
- turns on sitemaps using mezzanine's default implementation
- adds automated accessibility test tools (`pa11y-ci`) & configuration
- runs the tests in travis
- misc fixes across site to resolve accessibility issues
- adds some extra info about testing to the README, including js testing since it wasn't there either (my bad!)

notes:
- currently have `pa11y` set to ignore contrast errors: we can turn this off later by editing `.pa11yci.json`.
- the tests will crawl `/sitemap.xml` and use the additional paths in `.pa11yci.json`. right now they don't check the digitized work detail page or run a search on it because I don't think any are in the database to test with, but we could maybe create and load a fixture to do this?


ppa/archive/templates/archive/snippets/pagination.html
```diff
@@ -8,11 +8,11 @@
     <div class="page-controls">
             {# previous link at first (disabled if no previous)  #}
             {% if page_obj.has_previous %}
-            <a rel="prev" class="ui icon basic button" href="?{% querystring_replace page=page_obj.previous_page_number %}">
+            <a name="previous page" title="previous page" rel="prev" class="ui icon basic button" href="?{% querystring_replace page=page_obj.previous_page_number %}">
```
rlskoeser: does pa11y require both name and title?

ppa/archive/templates/archive/snippets/pagination.html
```diff
@@ -21,23 +21,23 @@
 
                 {#  always display current page, marked as active #}
                 {% if number == page_obj.number %}
-                    <a class="ui active basic button" href="?{% querystring_replace page=number %}">{{ number }}</a>
+                    <a title="page {{number}}" class="ui active basic button" href="?{% querystring_replace page=number %}">{{ number }}</a>
```
rlskoeser: not specific to your updates, but any thoughts on how to simplify the redundant code here without confusing the logic?

ppa/archive/templates/archive/snippets/pagination.html
```diff
@@ -8,11 +8,11 @@
     <div class="page-controls">
             {# previous link at first (disabled if no previous)  #}
             {% if page_obj.has_previous %}
-            <a rel="prev" class="ui icon basic button" href="?{% querystring_replace page=page_obj.previous_page_number %}">
+            <a name="previous page" title="previous page" rel="prev" class="ui icon basic button" href="?{% querystring_replace page=page_obj.previous_page_number %}">
                 <i class="chevron left icon"></i>
             </a>
             {% else %}
-            <a class="ui disabled icon basic button">
+            <a name="previous page" title="previous page" class="ui disabled icon basic button">
```
rlskoeser: perhaps disabled should be a `<span>` rather than an `<a>` ? seems misleading to add a title to something that is probably better ignored by a screen reader

ppa/archive/templates/archive/snippets/search_form.html
```diff
@@ -1,6 +1,7 @@
 {% load semanticui humanize %}
 <form class="ui form{% if error %} error{% endif %}">
     <fieldset class="ui vertical stackable grid segment">
+        <legend class="sr-only">text search fields</legend>
```
rlskoeser: I wonder if "keyword" would make more sense than "text" ?

ppa/common/fixtures/sites.json
```diff
@@ -0,0 +1 @@
+[{"model": "sites.site", "pk": 1, "fields": {"domain": "localhost:8000", "name": "Princeton Prosody Archive"}}]
```
rlskoeser: I suggest naming the fixture more specifically so it's more obvious what this is used for - test_site? ci_test_site?

test-requirements.txt
```diff
@@ -0,0 +1,3 @@
+pytest
```
rlskoeser: Do these requirements overlap with dev-requirements? We might want to think about restructuring our requirements files (concerned about redundancy; easy to make updates in one and not another, e.g. like the recent pytest version issue we ran into).  I'm ok with this for now, but would like to revisit soon (maybe before we adapt for another project).

README.rst
```diff
@@ -125,6 +125,23 @@ top-level project directory to your python path environment variable::
 Make sure you configure a test solr connection and set up an empty
 Solr core using the same instructions as for the development core.
 
+Javascript unit tests are written with `Jasmine <https://jasmine.github.io/>`_
+and run using `Karma <https://karma-runner.github.io/2.0/index.html>`_. To run
+them, you can use an ``npm`` command::
+
+    npm test
+
+Automated accessibility testing is also possible using `pa11y <https://github.com/pa11y/pa11y>`_
```
rlskoeser: Great, thanks for documenting these here

ppa/archive/templates/archive/snippets/pagination.html
thatbudakguy: yeah...I think it's because the link doesn't have an ID? I remember it prompting me to either add an ID or name even _after_ I added title in some cases; this was one. might be something to do with creating a unique handle for the link for the screen reader?

ppa/archive/templates/archive/snippets/pagination.html
thatbudakguy: keeping it as a link allows it to live with the other page links in the screen reader's link table, but still convey the information that it's disabled, which I think is still valuable to the user ("there's normally a link to go to the previous page here, but it's currently disabled") vs no info at all.

ppa/archive/templates/archive/snippets/search_form.html
thatbudakguy: i feel like some people think "keyword" is necessarily different from field scoped things like "title", "author" even though it's not. "text" is kinda vague though since it could mean "textual" (i.e. non-numeric) or the actual text of the pages in the PPA ("full text")....

@bwhicks do you have thoughts?

test-requirements.txt
thatbudakguy: hmm, that's a good point. I'm not sure what to call/how to think of those shared requirements, but it's definitely true that there's duplication we could avoid.

ppa/archive/templates/archive/snippets/pagination.html
thatbudakguy: I guess on a basic level we could just abstract a single button to a template that receives some information like if it's an arrow or a number, what number, what link it is, etc.

Then maybe do a little more work in the view to build a well-formed collection of button objects that we pass to that template and iterate it?

test-requirements.txt
rlskoeser: I've seen directories and includes used before, something like the structure described here: https://stackoverflow.com/questions/17803829/how-to-customize-a-requirements-txt-for-multiple-environments

This actually could simplify things - you ought to be able to install just one set of things and get everything you need.  We could still keep a top-level requirements.txt that includes the appropriate requirements/whatever.txt if that helps with deploy and travis-ci configuration.

We might want to shift to something like this now that we have a need for test requirements that exclude dev requirements. Not sure it should be in scope for this PR.

@bwhicks what do you think?

ppa/archive/templates/archive/snippets/search_form.html
rlskoeser: sounds fair to me, thanks for explaning

test-requirements.txt
bwhicks: @rlskoeser I've been having similar thoughts, especially now that our requirements are growing and getting a bit more complex. I wish python had something closer to npm in terms of fine-grained control.

I agree we should do it and agree that it's out of scope for this PR.

PR #114, opened 2018-10-12
Revise hathi import and indexing to include subtitles, sort titles, and more page details

rlskoeser:
-  Adds sort title and subtitle fields to `DigitizedWork`, makes them editable and visible in django admin, and populates them from Hathi Marc data.
- Use METS xml file to get more metadata for Hathi page images to include in Solr index, rather than relying solely on filenames

This doesn't include any logic for refining author handling or pulling in other creator information, but I think that should be a separate chore since the title and page index updates are blocking other features.


ppa/archive/models.py
```diff
@@ -119,9 +133,60 @@ def populate_from_bibdata(self, bibdata):
...
+            except ValueError:
+                # at least one record has a space here instead of a number
+                # probably a data error, but handle it
+                # - assuming no non-sort characters
+                non_sort = 0
+
+            self.sort_title = bibdata.marcxml.title()[non_sort:]
+
+            self.author = bibdata.marcxml.author() or ''
+            if '260' in bibdata.marcxml:
```
bwhicks: I think a note here for what `260` is in marc might be helpful down the line. Maybe.

ppa/archive/hathi.py
```diff
@@ -96,3 +120,51 @@ def marcxml(self):
...
+
+    @property
+    def text_file(self):
+        return METSFile(self.node.xpath('//m:file[@ID="%s"]' % self.text_file_id,
+                                        namespaces=self.ROOT_NAMESPACES)[0])
+    @property
+    def text_file_location(self):
+        return self.text_file.location
+
+      # <METS:div ORDER="1" LABEL="FRONT_COVER, IMAGE_ON_PAGE, IMPLICIT_PAGE_NUMBER" TYPE="page">
```
bwhicks: This is helpful in seeing what's going on, thanks for including it in the code.

ppa/archive/models.py
```diff
@@ -119,9 +139,66 @@ def populate_from_bibdata(self, bibdata):
...
+
+            # indicator 2 provides the number of characters to be
+            # skipped when sorting (could be 0)
+            try:
+                non_sort = int(bibdata.marcxml['245'].indicators[1])
+            except ValueError:
+                # at least one record has a space here instead of a number
+                # probably a data error, but handle it
+                # - assuming no non-sort characters
+                non_sort = 0
```
thatbudakguy: codecov thinks we haven't tested this case.

PR #120, opened 2018-10-17
Feature/hathi links

bwhicks:
* Adds template tag to render Hathi page links using the order attribute of solr page objects
* Adds and styled page reference via `page.label` per Zepplin
* Refactors tags for `page_url` and `page_image_url`

### DIFF
--- a/ppa/archive/models.py
+++ b/ppa/archive/models.py
@@ -346,16 +346,18 @@ class DigitizedWork(models.Model, Indexable):
 
                 pagefilename = os.path.join(self.hathi_content_dir, page.text_file_location)
                 with ht_zip.open(pagefilename) as pagefile:
-
-                    yield {
-                        'id': '%s.%s' % (self.source_id, page.text_file.sequence),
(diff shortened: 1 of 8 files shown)


thatbudakguy:
Here are some issues I'm seeing...unclear how things could be so different on my local dev, but posting anyway so we can debug them.

- page images are not loading for a large number of items. this looks like:
screen shot 2018-10-18 at 10 03 34 am which isn't ideal.
- page numbers aren't populating anywhere (see the above screenshot). this results in the empty `p.`
- when page images do load, they are of the correct _book_, but only ever of the cover. this looks like:
screen shot 2018-10-18 at 10 02 09 am you can see based on the image that it's the correct item.

When looking at the markup, I can see that both the link tags and image tags are actually pointing at different places:
screen shot 2018-10-18 at 10 06 53 am
screen shot 2018-10-18 at 10 06 47 am

However, both those image URLs resolve to the cover page image, and both those link URLs resolve to the cover page in hathi.

Issue #125, opened 2018-11-01
Page sometimes deformats when user presses back button in browser


mnaydan:
## testing notes
- steps to reproduce: 
  - load the archive search
  - make a change that updates results without reloading the page (any search or filter)
  - click on a link to navigate away
  - use browser back button to go back to archive search results

---

See screenshot. This seems to happen -only sometimes- when you do a keyword search, click on the title to get to the item detail page, and then press the back button. Do we need to build in a "Back" button within the site to return to the results page? 
<img width="1232" alt="screen shot 2018-11-01 at 11 41 02 am" src="https://user-images.githubusercontent.com/22945600/47862422-7670be80-ddcb-11e8-94d4-d283bb3010b7.png">

Edit: not only when you do a keyword search. I got this by filtering by collection, clicking on a title and then clicking back.

## dev notes
- [ ] send appropriate `vary` header (we have done this on winthrop)

mnaydan:
Did steps multiple times with various form changes and it seems to be fine now.

PR #133, opened 2018-11-07
Feature/add vary headers

bwhicks:
This request resolves the caching issue that produces #125 error by copying (and testing) the `VaryOnHeadersMixin` from `winthrop-django`.


rlskoeser:
Looks great.

PR #134, opened 2018-11-13
Feature/case insensitive sort

bwhicks:
This PR:

* Creates a case insensitive string field to use for `sort_title`
* Adds filters to `text_en` to provide more searching flexibility (following `winthrop-django`)
* Adds unit test coverage to new solr methods

### DIFF
--- a/DEPLOYNOTES.rst
+++ b/DEPLOYNOTES.rst
@@ -15,17 +15,23 @@ Deploy and Upgrade notes
   who were previously in the *Content Editor* group should be added
   to one of these, and the *Content Editor* group should be removed.
 
- * Solr schema changes are required for this update as well as a reindex of
-   both works and pages. Before these can be run, however, an updated
-   ``solrconfig.xml`` that includes the appropriate ``<lib/>`` declarations
-   must be installed on the core's `conf` folder. A correct copy may be found
-   in the `solr_conf` directory of this repository.
-   
-   After the Solr server has been restarted, the following commands should
-   be run::
+* Solr schema changes for this release require an updated ``solrconfig.xml``
+  with additional ``<lib/>`` declarations. A sample config file can
+  be found in ``ci/solrconfig.xml``. The following lines must be added::
 
+    <lib dir="${solr.install.dir:../../../..}/contrib/analysis-extras/lucene-libs/" regex="lucene-analyzers-icu-\d.*\.jar" />
+    <lib dir="${solr.install.dir:../../../..}/dist/" regex="solr-analysis-extras-\d.*\.jar" />
+    <lib dir="${solr.install.dir:../../../..}/contrib/analysis-extras/lib/" regex="icu4j-\d.*\.jar" />
+
+  Restart the Solr server to enable the new library paths.
+
+  Because this includes a Solr schema field type change that cannot be converted
+  automatically, the index must be cleared before changing the schema,
+  and then all content must be reindexed::
+
+    python manage.py index --clear all --index none
     python manage.py solr_schema
-    python manage.py index -c all -i all
+    python manage.py index
 
 0.9
 ---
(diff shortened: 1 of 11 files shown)


ppa/archive/solr.py
```diff
@@ -28,14 +28,63 @@ class SolrSchema(object):
...
+                "tokenizer": {
+                    # treat entire field as a single token
+                    "class": "solr.KeywordTokenizerFactory",
+                },
+                "filters": [
+                    # lower case to ensure alphabetical sort behaves as
+                    # expected
+                    {"class": "solr.LowerCaseFilterFactory"},
+                    # enable normalization without case folding
+                    # (preserve case for facets)
```
rlskoeser: If we're not using this field type for facets (which we shouldn't because we don't want them lower case), does this matter? We might be able to use the combined ICU filter that I think we're using for Winthrop text in that case. Not a huge difference, but might make this a little simpler.

bwhicks: Of course. I will gladly pull this out.

rlskoeser:
Please update the deploy notes to make sure it documents the steps needed for this solr update.

DEPLOYNOTES.rst
```diff
@@ -7,14 +7,29 @@ Deploy and Upgrade notes
...
 
 * Wagtail provides predefined groups for *Editor* and *Moderator*. Users
   who were previously in the *Content Editor* group should be added
   to one of these, and the *Content Editor* group should be removed.
 
+ * Solr schema changes are required for this update as well as a reindex of
+   both works and pages. Before these can be run, however, an updated
+   ``solrconfig.xml`` that includes the appropriate ``<lib/>`` declarations
+   must be installed on the core. A sample of the correct declarations can
+   be found in ``ci/solrconfig.xml``. The following lines must be added::
```
rlskoeser: Are we not providing a `solrconfig.xml` ? Or were we just using defaults before now?

Seems easier to me to provide a local version vs documenting what to add manually.

DEPLOYNOTES.rst
```diff
@@ -7,14 +7,29 @@ Deploy and Upgrade notes
...
+
+    <lib dir="${solr.install.dir:../../../..}/contrib/analysis-extras/lucene-libs/" regex="lucene-analyzers-icu-\d.*\.jar" />
+    <lib dir="${solr.install.dir:../../../..}/dist/" regex="solr-analysis-extras-\d.*\.jar" />
+    <lib dir="${solr.install.dir:../../../..}/contrib/analysis-extras/lib/" regex="icu4j-\d.*\.jar" />
+
+   After the Solr server has been restarted, the following commands should
+   be run::
+
+    python manage.py solr_schema
+    python manage.py index -c all -i all
```
rlskoeser: Why do we need to the `clear` option?

DEPLOYNOTES.rst
bwhicks: The `solrconfig.xml` could be used and copied over. However, I am reluctant to make the instruction to copy it over because of other defaults or changes that might be affected on the production core's `solrconfig' vs providing the `<lib/>` lines that can be added.

DEPLOYNOTES.rst
bwhicks: With the schema changes in the past, I have had issues without issuing the `-c all`. I waffled on whether I thought it was needed in this case. If you discover it isn't while testing locally, we should clearly drop it.

DEPLOYNOTES.rst
bwhicks: Would it be better to copy over what production's config looks like and make it our template going forward (perhaps as `.xml.sample` somewhere logical (`archive/fixtures` but seems off)?

rlskoeser: @bwhicks that sounds like a good solution to me. Maybe a top-level solr or solr_conf directory, something like we did for Winthrop? We might eventually want to customize other things too.

rlskoeser:
Now that I've finally gotten past the schema change issues, I'm able to confirm that this is working properly. @bwhicks please review my changes and instructions for the index and schema updates.

PR #179, opened 2018-11-28
Feature/analytics

bwhicks:
Adds flag `INCLUDE_ANALYTCICS` for settings.py and document usage in `local_settings.py.sample` and `DEPLOYNOTES`

N.B. Snippet for gtags is not compressed because it gains very little and complicates unit tests on Travis.


thatbudakguy:
we probably want to file this against `develop` rather than `master`, right?

rlskoeser:
@thatbudakguy thanks for catching, just edited it to switch the base to develop.

sitemedia/js/snippets/init_gtags.js
```diff
@@ -0,0 +1,8 @@
+/*
+  Google Analytics gtags snippet to be included if analytics are enabled.
+*/
+window.dataLayer = window.dataLayer || [];
+function gtag(){dataLayer.push(arguments);}
+gtag('js', new Date());
+
+gtag('config', 'UA-87887700-5');
```
rlskoeser: Since we're adding a local settings variable and setting it in the context anyway, what do you think of setting the UA code in the variable instead of hard-coding it in the javascript? Any benefit, or not something that should change?

thatbudakguy: In other words something like `GA_TRACKING_ID` which when set passes that variable into the js and when `None` turns off analytics? Sound good to me!

bwhicks: I thought about that and waffled. It probably shouldn't ever be changing outside of a release cycle, but I like the flexibility just in case. When I pick this back up, I'll refactor the js into the variable so we can change it if need be.

rlskoeser: Yes, if `None` or not set it would just turn off analytics.  I agree we don't really expect it to change, but I like having it in a config instead of code.

ppa/context_processors.py
```diff
@@ -9,6 +9,6 @@ def template_globals(request):
     context_extras = {
         'SHOW_TEST_WARNING': getattr(settings, 'SHOW_TEST_WARNING', False),
         'site': Site.objects.get_current(),
-        'INCLUDE_ANALYTICS': getattr(settings, 'INCLUDE_ANALYTICS', False)
+        'GTAGS_ANALYTICS_ID': getattr(settings, 'GTAGS_ANALYTICS_ID', False)
```
rlskoeser: `None` might make more sense than `False` here, but that is a _very_ minor quibble.

rlskoeser:
Changes look good to me.

PR #184, opened 2018-11-29
Feature/editorial list styles

thatbudakguy:


thatbudakguy:
pa11y is failing because I'm including the pagination template but there's no paginator data to populate it, which means some link elements are empty. I'll comment it out for now.

Issue #148, opened 2018-11-19
As a user, I want to exclude or include items from any of the collections in PPA so that I can refine my search to include relevant items.


thatbudakguy:
## Notes for testing
- In django admin, collections should now have a checkbox to control whether they are excluded by default from the public search; you should be able to edit this on the collection list and the individual collection edit page, and it should default to unchecked (not excluded)
- On the search page, collection filters should be pre-selected based on the exclude flag you set in the admin
- On the search page, selecting and deselecting collections should filter your results so that items in *any* of the collections selected are returned. If no collections are selected, you should get no results.
- The search page should include an 'Uncategorized' collection if there any items not in a collection. It should work the same way the other collection filters do.
- The 'Uncategorized' label should not be displayed on individual search results
- If all items are in a collection, you should not see an `Uncategorized` collection. (You can test this by creating a new collection and using the admin search to add everything to that collection.)

## dev notes
- [x] add exclude flag to collection model and make editable in admin
- [x] preselect collections on search form based on exclude flag
- [x] revise vue.js form logic to submit `collection=` when no collections are selected, to differentiate defaults from none selected
- [x] add logic to find items not in collection and provide an 'uncategorized' collection option

rlskoeser:
@mnaydan I think we should add a field to the collections in the admin so that you can configure which collections should be excluded by default. Does that sound right to you?

mnaydan:
@rlskoeser it would just be the "Dictionaries" and "Word Lists" ones that should be excluded by default. If it is easier for you guys to just hard code that, that's fine; I don't foresee us needing the capacity to change that anytime soon. (Of course if it's really easy to add a field in admin, my philosophy is to give us the capacity to control it. Trying to balance that philosophy with prioritizing dev work, though--I'll let you decide.)

mnaydan:
I checked everything in the testing notes and it's all working correctly, except for the attrition of 3 works: The Archive starts out with 5,007 works. If you deselect all the collections, and then reselect them all, you get 5,004 results total.

rlskoeser:
Figured out what's going on with the 5,007 vs 5,004. There are three items with no publication date (uc1.c2608792, uc1.b3011277, uc1.b2900825), and the change to use actual numbers in the date range part of the form instead of placeholders means that dates are now submitted any time the form is changed. Not sure what the best solution is; I wonder if we should go back to placeholders? (Maybe add logic to not submit them, but that doesn't seem ideal).

[...]

PR #185, opened 2018-11-29
Revise search collection logic to exclude admin-selected collections by default #148

rlskoeser:
- adds an exclude flag to `Collection` and makes it admin editable
- pre-selects collections on the search form based on the exclude flag
- sets a default collection value in Solr for works without a collection
- extends the collection field to handle a non-pk value to filter on the unset collection
- adjust vue.js form logic to send `collection=` when no collections are selected, so the view can differentiate between defaults and no collections


bwhicks:
This looks great and tests out locally. My only comment would be that adding a flag not to display `uncategorized` in the results list would probably look better and be less visually cluttered (since there will at least initially be quite a few books in that 'category'?)

PR #190, opened 2018-12-03
Add and order multiple authors on an editorial

bwhicks:
This PR:
* Adds a Wagtail `Person` snippet and configures it to be add-able via `StreamField`
* Adds the authors and date to the templates for editorial pages (both editorial proper and index)
* Add a link to authors with urls associated with them on the editorial page, but not on the editorial index page and order them.
* Does *not* address styling issues.


ppa/editorial/models.py
```diff
@@ -83,15 +81,34 @@ def route(self, request, path_components):
             return super().route(request, path_components)
 
 
+@register_snippet
+class Person(models.Model):
```
rlskoeser: I see why you added these under `editorial`, but we might want to use them for the contributor page at some point, so I think they ought to be under `pages`.  What do you think? 

(I had that in mind, but I didn't tell you that or document anywhere - I apologize).

ppa/editorial/models.py
```diff
@@ -83,15 +81,34 @@ def route(self, request, path_components):
...
+    name = models.CharField(max_length=255)
+    url = models.URLField(blank=True, default='')
```
rlskoeser: Let's add help text please. Maybe something along these lines?
- full name for the person as it should appear in the author list
- Personal website, profile page, or social media profile page for this person.

(The `URL` should be something we can use as a URI for the person when we're able to generate schema.org metadata for blogposts; there may be a better way to articulate that here in the help text.)

ppa/pages/templates/pages/content_page.html
```diff
@@ -26,6 +27,8 @@
 <article class="ui text container">
     <h1 class="ui header">
         {% firstof page.title page_title object.title %}
+        {% block attribution %}{% endblock %}
```
rlskoeser: This seems a little weird to me, but I don't feel strongly about it. Could make more sense to wrap the header contents in a `header` block that can be extend with `super`.

ppa/editorial/templates/editorial/snippets/attribution.html
```diff
@@ -0,0 +1,10 @@
+{# Snippet to include authors and date attribution to editorial posts #}
+{% for block in page.authors %}
+    {% if block.value.url and add_link %}
```
rlskoeser: Could do what we've done elsewhere when we sometimes have a link and sometimes not: https://github.com/Princeton-CDH/derrida-django/blob/master/derrida/books/templates/components/books-list-item.html#L25-L26 

I think that might be a bit better than this.

ppa/editorial/templates/editorial/snippets/attribution.html
```diff
@@ -0,0 +1,10 @@
...
+        <a href="{{ block.value.url }}" target="_blank">
+    {% endif %}
+        By {{ block.value.name }}
```
rlskoeser: This is displaying "By " for every author, not just the first.

ppa/editorial/templates/editorial/snippets/attribution.html
```diff
@@ -0,0 +1,10 @@
...
+        {% if block.value.url and add_link %}</a>{% endif %}
+        {% if not forloop.last %}, {% endif %}
```
rlskoeser: The newline is rendering as a space between author name and comma when this displays.

ppa/editorial/tests.py
```diff
@@ -89,9 +90,23 @@ def test_routing(self):
...
+        # NOTE: Should these be refactored into separate template
+        # checking test?
         self.assertContains(response, index_page.title)
         self.assertContains(response, editorial_page.title)
         self.assertContains(response, editorial_page.relative_url(site))
+        # each should be shown once
+        person_a = Person.objects.get(name='Person A')
+        person_b = Person.objects.get(name='Person B')
+        self.assertContains(response, person_a.name, count=1)
+        self.assertContains(response, person_b.name, count=1)
```
rlskoeser: Maybe check that "By " is only displayed once, as part of fixing that logic for multiple authors.

ppa/editorial/models.py
rlskoeser: Please add a brief docstring documenting what this is and where it's currently used / why it's here.

ppa/editorial/templates/editorial/snippets/attribution.html
```diff
@@ -0,0 +1,10 @@
...
+{% endfor %}
+/ On {{ page.first_published_at|date:"F j, Y"}}
```
rlskoeser: Authors are optional, but the `/` displays no matter what.

ppa/pages/templates/pages/content_page.html
bwhicks: This was actually so that the attribution line could fit in the `h1` header as as to be above the horizontal rule bar. Wrapping it would make inserting there hard, unless I'm missing something (totally likely)?

ppa/editorial/templates/editorial/snippets/attribution.html
bwhicks: Will fix and test. Woops.

ppa/editorial/templates/editorial/snippets/attribution.html
bwhicks: Fixing, too, obviously.

ppa/pages/templates/pages/content_page.html
rlskoeser: Oh, I didn't catch that. The attribution shouldn't be inside the `<h1>`, should it? Just the title? Maybe we need a `<header>` or something. But I don't feel that strongly about it; if the markup is provisional, we can leave for @thatbudakguy to sort out when he finishes the styles.

thatbudakguy: @bwhicks is exactly right; it's weird to put the attribution inside the `<h1>` but I do it so that I can extend the border below it to the correct length. this might all change because I didn't notice that the border lines are supposed to run past the margin on the right side, and I also didn't do them in a faded color because borders can't have opacity, so I might have to rework how they're rendered entirely.

ppa/editorial/templates/editorial/snippets/attribution.html
bwhicks: I had to do a hybrid of both approaches. The issue is that there's not a great way to use `yesno` and also have a switch for "never include" links. When I push I think what I did will be clearer and more explicit (if a bit more verbose in the name of template clarity)

ppa/editorial/templates/editorial/snippets/attribution.html
```diff
@@ -1,10 +1,17 @@
 {# Snippet to include authors and date attribution to editorial posts #}
-{% for block in page.authors %}
-    {% if block.value.url and add_link %}
-        <a href="{{ block.value.url }}" target="_blank">
+{% if page.authors %}
```
rlskoeser: @bwhicks I see what you mean, there are three different cases here which make the logic more complicated. Maybe you could use `firstof` to consolidate the cases somehow, but I'm not sure. I don't love the duplication, but it's not terrible and it's certainly readable. Works for me.

PR #195, opened 2018-12-05
Strip hyperlinks (and some other html) from description text for editorial pages

bwhicks:
This PR:

- Adds bleach as a dependency
- uses it to strip hyperlinks and other HTML other than its default safe set (minus `<a>`) from descriptions used in editorial index page.

### DIFF
--- a/ppa/pages/models.py
+++ b/ppa/pages/models.py
@@ -123,7 +123,8 @@ class PagePreviewDescriptionMixin(models.Model):
 
     # ('a' is omitted by subsetting and p is added to default ALLOWED_TAGS)
     #: allowed tags for bleach html stripping in description
-    allowed_tags = bleach.sanitizer.ALLOWED_TAGS[1:] + ['p']
+    allowed_tags = list((set(bleach.sanitizer.ALLOWED_TAGS) |
+                        set(['p'])) - set(['a']))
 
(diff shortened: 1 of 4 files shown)


ppa/pages/models.py
```diff
@@ -143,6 +148,11 @@ def get_description(self):
                     # break so we stop after the first instead of using last
                     break
```
rlskoeser: might be worth a comment here explaining why we need it

ppa/pages/models.py
```diff
@@ -120,6 +121,10 @@ class PagePreviewDescriptionMixin(models.Model):
     #: maximum length for description to be displayed
     max_length = 250
 
+    # ('a' is omitted by subsetting and p is added to default ALLOWED_TAGS)
+    #: allowed tags for bleach html stripping in description
+    allowed_tags = bleach.sanitizer.ALLOWED_TAGS[1:] + ['p']
```
rlskoeser: Is this relying on current order in the allowed tag list? I'd be more comfortable with explicitly filtering it out by tag name. Maybe set logic would be a good solution? (Or maybe it's too verbose)
`list((set(bleach.sanitizer.ALLOWED_TAGS) | set(['p'])) - set(['a']))`

PR #213, opened 2018-12-13
Support manual metadata entry for non-hathi digitized works

rlskoeser:
- updates digitized work with a `source` field to distinguish between hathi content and other content
- adjust fields that are displayed and edited in django admin; set hathi read only fields based on source 
- add checks to suppress logic, page index data, and marc metadata to skip non-hathi records
- customize detail view for non-hathi records, including suppressing search within the volume

Currently I have the alternate source set to **Other**; I will check with project team and see if this can be made more specific.

The only thing that isn't working right now is unapi for non-hathi records: we need to determine if there is a source where we can get MARC metadata (partly depends on the source question also). Might be possible for oclc records, but looks like it requires an API key - so making the metadata available for download (which we effectively do via unapi to generate the citation) is probably not within what we're allowed to do. Might need to generate simple dublin core or other metadata for Zotero for records where we can't get MARC metadata.


ppa/archive/admin.py
```diff
@@ -12,23 +12,38 @@ class DigitizedWorkAdmin(admin.ModelAdmin):
...
         'source_id', 'title', 'subtitle', 'author', 'enumcron', 'pub_date',
         'publisher', 'public_notes', 'notes', 'record_id'
     )
     filter_horizontal = ('collections',)
     # date_hierarchy = 'added'  # is this useful?
-    list_filter = ['collections', 'status']
+    list_filter = ['collections', 'status', 'source']
     actions = ['bulk_add_collection']
 
+    def get_readonly_fields(self, request, obj=None):
```
bwhicks: The logic behind this is such that if a work is accidentally made HATHI, those fields are no longer editable -- including the `source` field. At which point, the only way around this would be to delete it.

I confirmed this by trying it out, and that makes sense to me, but I want to make sure that's the intended functionality.

sitemedia/js/search.js
```diff
@@ -113,6 +113,8 @@ $(function(){
     }
 
     function onPageLoad() {
+        // remove no-js styles for CSS
```
bwhicks: This makes sense, though maybe `remove no-js styles for CSS now that Javascript has been detected`?

Issue #183, opened 2018-11-29
As an admin, I want to correct basic item-level metadata errors and preserve those corrections so that I can override discrepancies in source materials for display on the site.


mnaydan:
## Testing notes

To test this, please: 

- edit one of the following fields (or better yet several) in a record or two in the admin: `title, subtitle, sort_title, enumcron, author, pub_place, publisher, pub_date` for a digitized work and save it.
- the fields should be listed under `protected_fields` in the detail view in the admin
- Let the development team  know the source ids of the items that you have updated. The team will run the `hathi_import` and you can then verify that the fields that were modified were not updated or overwritten.

rlskoeser:
@mnaydan have you and your team made any edits in the database that we will need to preserve? If so, this becomes a 5 instead of a 3 so we can account for those.

mnaydan:
@rlskoeser we haven't--we just have a list of the things that need to be corrected.

rlskoeser:
@mnaydan great, thank you for confirming.

mnaydan:
Made edits (various combinations of deletions, additions, and edits) to title, subtitle, sort_title, enumcron, author, publisher, pub_date fields for hvd.hwg11q and to pub_place and enumcron for hvd.32044015547375. The correct fields were listed under protected fields and all edits were preserved after @bwhicks ran import script. Yay!

PR #260, opened 2019-02-22
Preserve content edits from automatic update

bwhicks:
This PR addresses issue #183 and allows content editors edits to be preserved from Hathitrust (or other) updates by setting a `protected_fields` flags field. 

- sets flags for any updated fields in those included in basic metadata that are user editable (i.e. collections are not part of this) when an item is saved in the admin
-  displays those flags as non-editable, human readable text in the change view. 
- prevents the `populate_from_bibdata` method of `DigitizedWork` from updating those fields, when they have been flagged.

The implementation makes `protected_fields` unconditionally return an instance of `ProtectedFlags` and explicitly makes the field non-nullable, which avoids `None` handling in custom field methods. It also allows the default to always be `ProtectedFlags.no_flags` (i.e. `0`).


ppa/archive/admin.py
```diff
@@ -59,6 +60,30 @@ def source_link(self, obj):
     source_link.admin_order_field = 'source_id'
     source_link.allow_tags = True
 
+    def save_model(self, request, obj, form, change):
+        '''Note any fields in the protected list that have been changed
+        and preserve in database.'''
+        # new object, created from scratch, nothing to track and preserve
+        # NOTE: Is this logic correct? Or should all the fields be preserved?
```
rlskoeser: This logic seems right to me - we don't expect (or support) new records for HathiTrust items to be created from django admin.

However, it does occur to me that the protected field flags are hathi specific - I think this logic should only trigger when source=hathi.

ppa/archive/admin.py
```diff
@@ -59,6 +60,30 @@ def source_link(self, obj):
...
+        if not change:
+            super().save_model(request, obj, form, change)
+            return
+        db_obj = DigitizedWork.objects.get(pk=obj.pk)
```
rlskoeser: `DigitizedWork` extends `TrackChangesModel` - is there any reason we can't use the `has_changed` method here?

ppa/archive/models.py
```diff
@@ -121,6 +122,76 @@ def stats():
         return stats
 
 
+class ProtectedFlags(Flags):
```
rlskoeser: These flags are very much specific to `DigitizedWork`, so I think the name should indicate that more clearly. I'm not sure how to do that without making it super long, though. `ProtectedWorkFields` maybe? Or are you using flags here to distinguish from the model field? `ProtectedWorkFieldFlags` ?

ppa/archive/models.py
```diff
@@ -184,6 +255,13 @@ class DigitizedWork(TrackChangesModel, Indexable):
     #: internal team notes, not displayed on the public facing site
     notes = models.TextField(blank=True, default='',
         help_text='Internal curation notes (not displayed on public site)')
+    #: metadata fields that should be preserved on bulk update because
+    # they have been modified by data editors.
+    protected_fields = ProtectedFlagsField(
+        default=ProtectedFlags,
+        help_text='Fields protected from bulk update because they have been '
+                  'manually edited.'
```
rlskoeser: I might make this even more specific - not just manually edited, but edited in django admin.

Should probably also mention that it's Hathi-specific.

ppa/archive/models.py
```diff
@@ -283,13 +361,16 @@ def populate_from_bibdata(self, bibdata):
         '''
         # store hathi record id
         self.record_id = bibdata.record_id
+        print(bool(bibdata.marcxml))
```
rlskoeser: looks like a leftover from debugging or testing something

ppa/archive/models.py
```diff
@@ -283,13 +361,16 @@ def populate_from_bibdata(self, bibdata):
...
 
         # set fields from marc if available, since it has more details
         if bibdata.marcxml:
+
             # set title and subtitle from marc if possible
             # - clean title: strip trailing space & slash and initial bracket
-            self.title = bibdata.marcxml['245']['a'].rstrip(' /') \
-                .lstrip('[')
+            if not self.protected_fields.title:
```
rlskoeser: I wonder if it would be cleaner to handle all the protected field logic in one place? Something like this: collect all the new bibliographic data in a dictionary, and then iterate over the dictionary data and set it on `self` unless the field is flagged as protected. 

I think it would make the code more readable and more efficient - handle the marc logic separate from the protected field logic.

In fact, it might be useful for the protected field logic to be in a separate method - that would make it possible to honor the protected flags if we do any other kind of bulk updates from scripts or console.

ppa/archive/tests/test_models.py
```diff
@@ -15,13 +15,51 @@
...
+        comparison = [
+            DigitizedWork._meta.get_field('enumcron').verbose_name,
+            DigitizedWork._meta.get_field('sort_title').verbose_name.capitalize(),
+            # no verbose name, so should be capitalized
+            DigitizedWork._meta.get_field('title').verbose_name.capitalize()
+        ]
+        # should be a comma-joined list of the verbose names of the
+        # fields, in alphabetical order; avoided hard coding in case
+        # verbose name changes
+        assert str(fields) == ', '.join(comparison)
```
rlskoeser: I find this test hard to read; I wonder if the verbose name logic should be split out so it's testable directly?

ppa/archive/models.py
```diff
@@ -121,6 +122,76 @@ def stats():
...
+        all_fields = list(self.__members__.keys())
+        return [field for field in all_fields
+                if getattr(self, field)]
+
+    def __str__(self):
+        protected_fields = self.as_list()
+        verbose_names = []
+        for field in protected_fields:
+            field = DigitizedWork._meta.get_field(field)
+            # if the field has no actual verbose name, just capitalize
```
rlskoeser: There isn't an existing method we can use here?

ppa/archive/models.py
```diff
@@ -121,6 +122,76 @@ def stats():
...
+    def all_fields(cls):
+        return list(ProtectedFlags.__members__.keys())
+
+    def as_list(self):
+        '''Return combined flags as a list of field names.'''
+        all_fields = list(self.__members__.keys())
+        return [field for field in all_fields
+                if getattr(self, field)]
+
+    def __str__(self):
```
rlskoeser: Can't you make use of the `to_simple_str` method on flags objects? Maybe just change `|` to `,`? I'm not sure how important it is to change internal name to verbose names, this is a backend interface. Maybe we can get feedback from the project team on that. 

If we do want verbose names, then should probably either iterate over set flags or maybe use simple str as the starting point.

ppa/archive/models.py
```diff
@@ -121,6 +122,76 @@ def stats():
...
+        '''Give Django information needed to make
+        :class:`ProtectedField.no_flags` default in migration.'''
+        # (import path, [args], kwargs)
+        return ('ppa.archive.models.ProtectedFlags', ['no_flags'], {})
+
+    @classmethod
+    def all_fields(cls):
+        return list(ProtectedFlags.__members__.keys())
+
+    def as_list(self):
```
rlskoeser: Can't you just use the existing flag `__iter__` method?

ppa/archive/models.py
```diff
@@ -121,6 +122,76 @@ def stats():
...
+
+    @classmethod
+    def deconstruct(cls):
+        '''Give Django information needed to make
+        :class:`ProtectedField.no_flags` default in migration.'''
+        # (import path, [args], kwargs)
+        return ('ppa.archive.models.ProtectedFlags', ['no_flags'], {})
+
+    @classmethod
+    def all_fields(cls):
```
rlskoeser: I thought you should just be able to iterate over the class to get the defined fields, but I see from experimentation that it returns the verbose name from the string method! I'm not sure that's actually the behavior we want?

ppa/archive/models.py
rlskoeser: Also, if we do decide we want verbose names - can we calculate them once and store rather than converting every time `str` is called?

ppa/archive/models.py
rlskoeser: I checked the code climate report, and it's complaining about the repeated/similar code where you're checking for protected fields, so refactoring would resolve that also.

ppa/archive/admin.py
bwhicks: I will add a flag for this re: hathi

ppa/archive/admin.py
bwhicks: I think I have a way to use this using `refresh_from_db`; the wrinkle as discussed being the obj is a new creation and not aware of the database. If that doesn't work, I'll note the problem.

ppa/archive/models.py
bwhicks: That was the use of `Flags` (distinguishing from the model field). I like the two proposed options.

bwhicks: Actually, I like `ProtectedWorkFieldFlags` and `ProtectedWorkField` because it's a `models.Field` subclass.

ppa/archive/tests/test_models.py
bwhicks: Rendered a non-issue by not needing the verbose fields

ppa/archive/models.py
bwhicks: Sadly, no. The `_meta` is how you get access to field names and they are pretty low-level. Thankfully no no longer an issue.

ppa/archive/models.py
bwhicks: Thanks for this re: `to_simple_str`.

ppa/archive/models.py
bwhicks: Likewise for this! Much simpler even if I have to do a single dot split (shouldn't be too punishing).

ppa/archive/models.py
bwhicks: I think we're done with since we're not doing verbose names.

ppa/archive/models.py
bwhicks: Thanks for the spot.

ppa/archive/models.py
bwhicks: Agreed. I've made it point more specifically to Hathi.

ppa/archive/admin.py
bwhicks: No dice on `has_changed` and `refresh_from_db`. In the end, cleanest way is to compare against a DB copy. I've added a note explaining why I'm not using `has_changed`.

ppa/archive/models.py
bwhicks: This worked out well. I was able to abstract the check out to a nice, short loop.

bwhicks:
@rlskoeser I believe this is ready for another pass. I refactored the protected_fields logic so as to take a dictionary of fields and values to set, simplified the logic for the flags significantly, and revised the migrations with appropriate class names, etc.

ppa/archive/models.py
```diff
@@ -136,50 +136,28 @@ class ProtectedFlags(Flags):
...
+        :class:`ProtectedWorkFieldFlags.no_flags` default in migration.'''
         # (import path, [args], kwargs)
-        return ('ppa.archive.models.ProtectedFlags', ['no_flags'], {})
+        return ('ppa.archive.models.ProtectedWorkFieldFlags', ['no_flags'], {})
 
-    @classmethod
-    def all_fields(cls):
-        return list(ProtectedFlags.__members__.keys())
+    def __str__(self):
+        return ', '.join(sorted(self.to_simple_str().split('|')))
```
rlskoeser: I think I would have just done a string replace here, rather than splitting and sorting, since it looks like the flags are returned in the order they are defined on the flag subclass (which is a logical order, since you defined it). But this is fine.

rlskoeser:
Updates look good to me.

PR #281, opened 2019-03-28
Add new HathiTrust items from Django admin

rlskoeser:
- refactors logic from `hathi_add` script into new `ppa.archive.util.HathiImporter` for reuse
- reports what was done in human-readable fashion, with links to admin and public views for existing ids that were skipped or newly created records
- new `AddFromHathi` form and view, bound to admin urls for `DigitizedWork`
- use current logged in user for log entries created to document the record creation

### DIFF
--- a/ppa/archive/util.py
+++ b/ppa/archive/util.py
@@ -1,0 +1,125 @@
+'''
+
+Utility code related to :mod:`ppa.archives`
+
+'''
+
+from collections import OrderedDict
+from json.decoder import JSONDecodeError
+
+from ppa.archive.hathi import HathiItemNotFound, HathiItemForbidden
+from ppa.archive.models import DigitizedWork
+from ppa.archive.signals import IndexableSignalHandler
+from ppa.archive.solr import get_solr_connection
+
+
+class HathiImporter:
+    '''Logic for creating new :class:`~ppa.archive.models.DigitizedWork`
+    records from HathiTrust. For use in views and manage commands.'''
+
+    existing_ids = None
+
+    #: status - successfully imported record
+    SUCCESS = 1
+    #: status - skipped because already in the database
+    SKIPPED = 2
+
+    #: human-readable message to display for result status
+    status_message = {
+        SUCCESS: 'Success',
+        SKIPPED: 'Skipped; already in the database',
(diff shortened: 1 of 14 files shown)


rlskoeser:
@bwhicks as we discussed, I kept the refactor to the `hathi_add` script & view only. I think the `hathi_import` script could be refactored at some point too - shift the logic into the importer class and combine the two scripts - but as you suggested, it was too much to tackle now.  

I did come up with a helpful way of framing it that might not be in the code anywhere, which is a "local" import (pairtree data already available) and a "remote" import (data must be fetched.)

ppa/archive/util.py
```diff
@@ -0,0 +1,136 @@
...
+    SKIPPED = 2
+
+    #: human-readable message to display for result status
+    status_message = {
+        SUCCESS: 'Success',
+        SKIPPED: 'Skipped; already in the database',
+        hathi.HathiItemNotFound: 'Error loading record; check that id is valid.',
+        hathi.HathiItemForbidden: 'Permission denied to download data.',
+        # only saw this one on day, but this was what it was
+        JSONDecodeError: 'HathiTrust catalog temporarily unavailable.'
```
bwhicks: I think this error might be worth making more specific (for when a user reports a problem to us): i.e., `HathiTrust catalog temporarily unavailable, malformed response` as a flag so that we know the JSON was bad in some way?

ppa/archive/util.py
```diff
@@ -0,0 +1,136 @@
...
+        '''Check for any ids that are in the database so they can
+        be skipped for import.  Populates :attr:`existing_ids`
+        with an :class:`~collections.OrderedDict` of htid -> id for
+        ids already in the database and filters :attr:`htids`.
+
+        :param htids: list of HathiTrust Identifiers (correspending to
+            :attr:`~ppa.archive.models.DigitizedWork.source_id`)
+        :returns: tuple of:
+            - list of htids not present in the database
+            - dictionary of
```
bwhicks: `dictionary of` looks like it's missing part of a phrase?

rlskoeser:
Thanks for the feedback @bwhicks, I like your suggested wording and appreciate the catch on the missing docs.

PR #287, opened 2019-04-04
Refactor/reactive search

thatbudakguy:


srcmedia/js/index.js
```diff
@@ -1,6 +1,8 @@
 import PitBar from './pitbar'
 import NavMenu from './modules/NavMenu'
 
+document.firstElementChild.classList.remove('no-js') // remove the no-js class
```
rlskoeser: 👍

srcmedia/ts/lib/form.ts
```diff
@@ -0,0 +1,59 @@
...
+class RxForm extends Component implements Reactive<RxFormState> {
+    element: HTMLFormElement
+    state: Subject<RxFormState>
+    target: string
+
+    constructor(element: HTMLFormElement) {
+        super(element)
+        this.state = new Subject()
+        // FIXME naively assuming that the form targets the current path
+        this.target = window.location.pathname
```
rlskoeser: could check for form action attribute and default to current url if not set

srcmedia/ts/lib/output.ts
```diff
@@ -0,0 +1,27 @@
+import { Subject } from 'rxjs'
+
+import { Component, Reactive } from '../lib/common'
+import { ImageLazyLoader } from '../../js/modules/LazyLoad'
+
+type RxOutputState = string
+
```
rlskoeser: Missing docstring here to document what this is. This is for updating the results?

srcmedia/ts/lib/form.ts
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Update the form's state. We don't need to change the DOM, so all we do
+     * is call next().
+     *
+     * @param {RxFormState} state
+     * @returns {Promise<void>}
+     * @memberof RxForm
+     */
+    async update(state: RxFormState): Promise<void> {
+        this.state.next(state)
```
rlskoeser: Should I know what calling `next()` does? Is there context I'm missing to understand this?

srcmedia/ts/components/PageSearchForm.ts
```diff
@@ -0,0 +1,42 @@
...
+     * Also updates the browser history, saving the search.
+     *
+     * @returns {Promise<void>}
+     * @memberof PageSearchForm
+     */
+    async submit(): Promise<void> {
+        return fetch(`${this.target}?${this.serialize()}`, ajax)
+            .then(res => res.text())
+            .then(html => this.update({ results: html }))
+            .then(() => window.history.pushState(null, 'PPA Archive Search', this.serialize()))
```
rlskoeser: I think the pushed state is missing the `?` in the url - when I tested running it, the ajax form works but when I press enter I get a bad url: `archive/mdp.39015009444814/query%3Dinsect/?query=insect`

Related: should we disable submitting the form on enter?

Additionally: we shouldn't hardcode `PPA Archive Search` in the history; can you grab the html title from the page?

thatbudakguy: I was running into this too! I thought it was some kind of encoding thing but I think you're right, just needs the `?`.

Makes sense to disable enter to submit & agree about grabbing the title, seems like a good idea.

srcmedia/ts/lib/form.ts
thatbudakguy: it's a method on all Observable-like datatypes (e.g. `Subject`) that pushes the next item into the sequence, analogous to `self.next()` on a python generator.

srcmedia/ts/lib/output.ts
thatbudakguy: yeah, for now it's just a "dumb" representation of the returned html as a string. I'll add a note

srcmedia/ts/lib/form.ts
thatbudakguy: I tried doing this but setting `action=` manually causes the form to fail when js is turned off, and I can't figure out how to get the embedded `action` property out of the Django form, since it seems like Django just "knows" it?

ppa/archive/forms.py
```diff
@@ -387,6 +390,36 @@ class SearchWithinWorkForm(forms.Form):
             '_align': 'left'
     }))
 
+    def as_json(self):
```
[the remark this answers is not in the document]
thatbudakguy: not anymore, although I considered reviving it re: the discussion on how to send the form's `action=` to the frontend, since Django knows it but I can't get it into the markup without breaking the form in some way...

srcmedia/js/modules/Utilities.js
```diff
@@ -0,0 +1,19 @@
+export const toArray = value => Array.isArray(value) ? value : value ? [value] : undefined
+
+export const ajax = {
+    headers: { // this header is needed to signal an ajax request to Django
+        'X-Requested-With': 'XMLHttpRequest',
+    }
+}
+export const parser = new DOMParser()
+
+export const reduceFacets = (acc, cur) => {
```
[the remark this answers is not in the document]
thatbudakguy: oh, these are from old implementations of the reactive search - good catch. Not used yet, but I put in since we may or may not need them when we start implementing the more complex text facet-type components.

srcmedia/ts/lib/form.ts
thatbudakguy: although I guess there's no harm in checking, just seems like django forms generally don't have `action=`

rlskoeser: We don't usually bother putting one when we're submitting to the same url that the form is on.  It's not a django-specific thing, just taking advantage of default browser behavior to submit to the current url if no action is set.

Thinking about it more, I'm not sure a reactive search would make sense for a form that submits to a different url... 

I'm ok with whichever way you want to handle it; fine if you just revise the comment to say that's current practice for django forms we want to be reactive.

Issue #289, opened 2019-04-08
As a content editor, I want to insert block quotes into the page so that I can use a special style to highlighted quoted material.


thatbudakguy:
## testing notes
- you can now insert a blockquote when editing a "paragraph" block on any content page (editorial, about, etc). the button looks like a quote:
<img width="360" alt="bq" src="https://user-images.githubusercontent.com/4924494/55819908-d6c90200-5ac7-11e9-9d3c-c5a6bcc3f4ec.png">

- you can either highlight some text you've already written and click the quote button, or click it and then start typing
- when published, the color of the border line on the left should match the color of the section you're in (editorial, about, etc).
### designs
[desktop](https://zpl.io/bJ73DNK)

thatbudakguy:
bumping to a 2 because we didn't have the actual blockquote functionality implemented; i copied the wagtail editor extension code from s&co.

thatbudakguy:
moving back to a 1 since I didn't actually _write_ any new code, just copy it over

mnaydan:
Block quote button works by highlighting or typing.
Color matches page section (tested in editorial and about).

Nick stripped the color side bar if block quote previews on tile on editorial list view; it'll just show up as italics now.

PR #290, opened 2019-04-09
Blockquotes (#289)

thatbudakguy:


ppa/pages/wagtail_hooks.py
```diff
@@ -0,0 +1,34 @@
+import wagtail.admin.rich_text.editors.draftail.features as draftail_features
+from wagtail.admin.rich_text.converters.html_to_contentstate import \
+    BlockElementHandler
+from wagtail.core import hooks
+
+# blockquote registration example taken from:
+# http://docs.wagtail.io/en/v2.4/advanced_topics/customisation/extending_draftail.html#creating-new-blocks
```
rlskoeser: Interesting! Thanks for including the link in the comments, very helpful

srcmedia/scss/_mixins.scss
```diff
@@ -180,6 +180,25 @@
             @include small-font-scale;
         }
     }
+
+    blockquote {
+        margin: 2rem 0 2rem 1.5rem;
+        padding-left: 1.5rem;
+        border-left-style: solid;
+        border-left-width: 5px;
+        border-left-color: $pig-pink;
```
rlskoeser: Seems a little weird to have the specific color here. I don't imagine there's an easy way to have an `$active-color` (or something) that is different in different sections of the site, but I do wonder about using a semantic name instead of the actual color - something like `$about-highlight` or `$editorial-highlight` (or whatever it is semantically, I don't know if it's a highlight color; theme color?).

thatbudakguy: yeah, I absolutely agree - when I refactor this I'll probably either add new variables exactly like that, or change the `@content-page()` mixin to take a color param (or both).

PR #295, opened 2019-04-11
Add last modified headers to archive and digitized work detail views

rlskoeser:


ppa/archive/views.py
```diff
@@ -52,8 +51,8 @@ class DigitizedWorkListView(AjaxTemplateMixin, ListView):
     # search title query field syntax
     search_title_query = '{!type=edismax qf=$search_title_qf pf=$search_title_pf v=$title_query}'
 
-
     def get_queryset(self, **kwargs):
+        print("*** get queryset")
```
thatbudakguy: should this still be in here?

rlskoeser: Ha, nope! I was testing whether the view is still getting called (it does) and had another thought about how we might avoid generating the whole response but decided to wait on that and forgot to take this back out. I'll clean it up.

ppa/archive/views.py
```diff
@@ -320,8 +319,35 @@ def get_context_data(self, **kwargs):
...
+    def last_modified(self):
+        '''override last modified logic to work with Solr'''
+
+        # override sort to return most recent modification date,
+        # only return last modified value and nothing else
+
+        # use most recent modification time of anything in the index,
+        # since any search on the archive page coud change based
+        # on work or text content changing in the index, (could add
+        # or remove results from any particular set)
```
thatbudakguy: wish there was a smarter way to do this, but it seems really hard, if not downright impossible.

rlskoeser: Agreed. I know in the past I've used the most recently modified date of anything in the current result set, but that doesn't account for items that might previously have been in that result list and aren't anymore. This still may not handle deletions - although with our suppression logic in PPA it should work pretty well.  And even though it might be more generous than necessary, it's still a lot better than nothing.

PR #324, opened 2020-01-09
Upgrade to django 2.2 and wagtail 2.7

rlskoeser:
Wagtail upgrade was required because wagtail 2.4 does not work with django 2.2

- updated custom admin actions to check the proper permissions, to make sure that things will work appropriately with django's new view permission
- removed our local `json` template tag in favor of the new `json_script` 
- other minor cleanup as needed

I was getting a lot of `ResourceWarning` warnings about a socket not being closed when I ran the tests; when I looked this up, it seems to be caused by a requests session not getting closed. I made a change to SolrClient that improved this. There might still be another session somewhere not getting closed, but I'm not sure where (I see three warnings for unclosed sockets when I run the tests with `pytest -W once`).


ppa/archive/templates/archive/snippets/results_list.html
```diff
@@ -1,15 +1,13 @@
...
 via ajax, e.g. to update the search results.
 {% endcomment %}
 <div class="data">
     <pre class="count">
         Displaying {{ paginator.count|intcomma }} digitized work{{ paginator.count|pluralize }}
     </pre>
-    <pre class="facets">
-        {{ facet_ranges.pub_date|json }}
-    </pre>
+    {{ facet_ranges.pub_date|json_script:"facets" }}
```
thatbudakguy: love that `json_script` lets you specify the id of the output `<script>` tag!

ppa/archive/tests/test_templatetags.py
```diff
@@ -89,14 +89,3 @@ def test_solr_highlight():
...
-
-def test_json():
-    # should be the same as json_dumps()
-    val = {
-        'a': 'b',
-        'c': 'd',
-        'e': 3,
-        'f': ['g', 6, 'h']
-    }
-    assert json_dumps(val) == mark_safe(json.dumps(val))
```
thatbudakguy: nice, hopefully less warnings about `mark_safe()` now

requirements.txt
```diff
@@ -1,9 +1,9 @@
-django>=1.11.16,<2.0
+django>=2.2,<3.0
```
thatbudakguy: we could go up to 3 if we wanted now, right?

srcmedia/js/search.js
```diff
@@ -57,7 +57,7 @@ $(function(){
         $('.workscount').addClass('loading') // turn on the loader
         req.then(res => res.text()).then(html => { // submit the form and get html back
             $$paginationTop.html($(html).find('.page-controls').html()) // update the top pagination
-            updateHistogram(JSON.parse($(html).find('pre.facets').html())) // update the histogram
+            updateHistogram(JSON.parse($(html).find('script#facets').html())) // update the histogram
```
thatbudakguy: nice to not have this in a `<pre>` tag anymore!

requirements.txt
rlskoeser: Maybe, but it just came out in December and our current LTSA says we'll stick to LTS releases. I'm thinking we should stick to 2.2 for now.

Issue #274, opened 2019-03-13
keyword search stemming is too aggressive


rlskoeser:
We should index page content in both stemmed and unstemmed versions, with boosting on the unstemmed version so that unstemmed matches will be prioritized.

Notes from @mm4CDH :
> Type the word "decadence" and you get results for "decade." That's the tricky one. It's only keyword search where it's a problem.
> It's mostly a problem for reverse stemming, shorter words that live inside big words but you want to search for the short word.

From @mnaydan:
> I remember running into this when I was testing--and might be my fault for not thinking about keyword search implications at the time. Title search seems to be working the best (but still isn't 100% perfect--"pronunciation" returns mostly "pronunciation" for the first ~50 results but we get some "pronunciations" too after that, before more "pronunciation"). "iamb" in keyword search returns "iambe" "iamb" "iambs" all mixed.

When I checked with PUL devs, they said:
> We usually index in both “ssim” (stands for string/stored/indexed/multiple) and “tesim” (tokenized/english/string/indexed/multiple) for that reason
> Although I guess if you’re searching within you probably want it tokenized. :disappointed:

For the unstemmed version, we should consider using a very simple regex tokenization (maybe [Simplified Regular Expression Pattern Tokenizer](https://lucene.apache.org/solr/guide/6_6/tokenizers.html#Tokenizers-SimplifiedRegularExpressionPatternTokenizer) that doesn't throw away all punctuation (which the default tokenization does), because it would be valuable to search on punctuation like `/` or `|` that is used to mark meter. This version of the index may also better support searching for things like AB rhyme scheme notation (since A normally gets thrown out as a stop word).

mnaydan:
The weirdest thing is that some words yield blank results! (e.g., the first 9 results are blank for "decadence" and the first 78 results are blank for "grass"). 

<img width="866" alt="Screen Shot 2020-04-27 at 7 15 15 AM" src="https://user-images.githubusercontent.com/22945600/80368691-1ee5ce80-885b-11ea-8d25-2c993c4aa129.png">

The boosting for the unstemmed words seems to be way better (yay!) but still isn't perfect? (e.g., result 9 for search term "iamb" is "iambe" even though there are more results that are just "iamb"; and when dictionaries and wordlists are included the first result is actually "lambs" (like a baby sheep - OCR error but still, the plural appears first); result 2 for "wheel" is "wheels" and "wheeling"). This might just be as good as it gets, which is pretty good and definitely better than what it was!

[...]

Issue #273, opened 2019-03-08
Installation notes


vineetbansal:
Hello,

I've been able to install/run the ppa application on Windows, but at certain points in the process had to work some installation issues that may be worth your attention. Sorry for dumping all this in one thread - some of these are not really 'issues' but just notes, some may be common knowledge to people working with django (and thus weren't mentioned in the README) but weren't obvious to me, and finally some may be specific to my case.

  - the README explicitly mentions Python 3.5 but everything worked fine for my Python 3.6.3 virtuenenv (except for one minor issue at the end of this list).
  - I had to make up a secret key in local_settings.py before running any `manage` command.
  - The logging mechanism in local_settings.py expects a handler called `console` but this is missing in the `handlers` section. I added something like:
```
        'console': {
            'class': 'logging.StreamHandler',
            'level': 'DEBUG',
            'stream': 'ext://sys.stdout'
        }
```

  - The README makes it seem like one can choose any names for cores/collections (`..using the core/collection names for development and testing that you configured in local settings..`) but really one has no choice in the matter, since `ppa/archive/solr.py` is hardwired to expect `default` as a key. The tests are wired to expect the `test` key (not very sure of this though). The example using `SOLR_CORE` and `SOLR_TEST_CORE` may be confusing here too.

  - On running `manage.py solr_schema`, I found that the code tries to make a connection to
```
http://127.0.0.1:8983/solr/ppa/schema/fieldtypes?wt=json&...
```
i.e. using the `collection` key value in the URL (`ppa`). However, on my SOLR installation (version 7.7.0), this caused a 404, and the correct URL seems to be:
```
http://127.0.0.1:8983/solr/default/schema/fieldtypes?wt=json&...
```
i.e. it uses the `core` name in the URL.
This was also a problem when running the unit tests later. I got around these by simply making sure the collection name matched the core name for both keys in SOLR_CONNECTIONS
```
    'default': {
        'COLLECTION': 'default',  # changed from 'ppa'
    ...
```

  - For the npm stuff, I had to do a `npm install gulp` before anything else.

  - At some point, I remember `manage.py migrate` giving me message like `Your models have changes that are not yet reflected in a migration, and so won't be applied. Run 'manage.py makemigrations' to make new migrations ...`, so I did just that to keep it happy before running migrate.

[...]

PR #333, opened 2020-04-21
solr stemming and readme

kmcelwee:
Issue #274 addressed by changing solr stemming. Issue #273 addressed by updating readme and adding a troubleshooting section.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -42,8 +42,8 @@ Initial setup and installation:
 
 - **recommended:** create and activate a python 3.6 virtualenv::
 
-     virtualenv ppa -p python3.6
-     source ppa/bin/activate
+     virtualenv venv -p python3.6
+     source venv/bin/activate
 
 - Use pip to install required python dependencies::
 
@@ -62,8 +62,8 @@ Initial setup and installation:
   solr config, using the core/collection names for development and
   testing that you configured in local settings::
 
-    solr create -c SOLR_CORE -d solr_conf
-    solr create -c SOLR_TEST_CORE -d solr_conf
+    solr create -c ppa -d solr_conf
+    solr create -c ppa-test -d solr_conf
 
 - Run the manage command to configure the schema::
 
@@ -78,7 +78,7 @@ Initial setup and installation:
 
     python manage.py hathi_import
 
-- Then index the imported content into Solr:
+- Then index the imported content into Solr::
 
     python manage.py index
 
@@ -90,6 +90,8 @@ before running the server.
 
 - To build all styles and js for production, including semantic UI::
 
+    npm install
+    npm install -D
     npm run build
 
 Alternatively, you can rebuild just the custom files or semantic independently.
@@ -188,6 +190,17 @@ to the ``docs`` folder in your worktree::
 In the ``sphinx-docs`` folder, use ``make docs`` to build the HTML documents
 and static assets, add it to the docs folder, and commit it for publication on
 Github Pages. After the build completes, push to GitHub from the ``docs`` folder.
+
+Troubleshooting
+-------------
+
+Trouble launching Solr? Try configuring Java to version 8. Later versions of 
+Java will cause Solr 6 to time out.
+
+Solr changes not reflected in search results? ``solrconfig.xml`` must be 
+updated in Solr's main directory: ``solr/server/solr/[CORE]/conf/solrconfig.xml``
+
+We currently do not support Python 3.7 or Solr 7 and above.
 
 License
 -------
--- a/README.rst
+++ b/README.rst
@@ -33,17 +33,15 @@ This repo uses `git-flow <https://github.com/nvie/gitflow>`_ conventions; **mast
 contains the most recent release, and work in progress will be on the **develop** branch.
 Pull requests should be made against develop.
 
-Python 3.6 / Django 1.11 / Node 10.5.0 / MariaDB (MySQL) 5.5 w/ timezone info
+Python 3.6 / Django 2.2 / Node 10.5.0 / MariaDB (MySQL) 5.5 w/ timezone info / Solr 6.6 (requires Java 8)
 
 Development instructions
 ------------------------
 
 Initial setup and installation:
 
-- **recommended:** create and activate a python 3.6 virtualenv::
-
-     virtualenv venv -p python3.6
-     source venv/bin/activate
+- **recommended:** create and activate a python 3.6 virtual environment, perhaps with 
+`virtualenv` or `venv`
 
 - Use pip to install required python dependencies::
 
@@ -54,7 +52,7 @@ Initial setup and installation:
 
    cp ppa/local_settings.py.sample ppa/local_settings.py
 
-- Create a database, configure in local settings, and run migrations::
+- Create a database, configure in local settings in the `DATABASES` dictionary, change `SECRET_KEY`, and run migrations::
 
     python manage.py migrate
 
@@ -91,7 +89,6 @@ before running the server.
 - To build all styles and js for production, including semantic UI::
 
     npm install
-    npm install -D
     npm run build
 
 Alternatively, you can rebuild just the custom files or semantic independently.
@@ -190,17 +187,6 @@ to the ``docs`` folder in your worktree::
 In the ``sphinx-docs`` folder, use ``make docs`` to build the HTML documents
 and static assets, add it to the docs folder, and commit it for publication on
 Github Pages. After the build completes, push to GitHub from the ``docs`` folder.
-
-Troubleshooting
--------------
-
-Trouble launching Solr? Try configuring Java to version 8. Later versions of 
-Java will cause Solr 6 to time out.
-
-Solr changes not reflected in search results? ``solrconfig.xml`` must be 
-updated in Solr's main directory: ``solr/server/solr/[CORE]/conf/solrconfig.xml``
-
-We currently do not support Python 3.7 or Solr 7 and above.
 
 License
 -------
--- a/README.rst
+++ b/README.rst
@@ -40,8 +40,7 @@ Development instructions
 
 Initial setup and installation:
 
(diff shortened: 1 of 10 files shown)


README.rst
```diff
@@ -42,8 +42,8 @@ Initial setup and installation:
 
 - **recommended:** create and activate a python 3.6 virtualenv::
 
-     virtualenv ppa -p python3.6
-     source ppa/bin/activate
+     virtualenv venv -p python3.6
```
kmcelwee, line 45: the venv "ppa" placed lib/bin into the django directory ppa. This caused many problems.

ppa/local_settings.py.sample
```diff
@@ -18,7 +18,7 @@ ALLOWED_HOSTS = []
 
 # SECURITY WARNING: keep the secret key used in production secret!
 # http://www.miniwebtool.com/django-secret-key-generator/
-SECRET_KEY = ''
+SECRET_KEY = '123456' # CHANGE ME !
```
kmcelwee, line 21: This tripped me up for a few minutes and was a concern in #273. This isn't a security concern, correct?

ppa/local_settings.py.sample
```diff
@@ -94,29 +94,29 @@ STATIC_ROOT = os.path.join(BASE_DIR, STATIC_URL.strip("/"))
...
+#             'level': 'ERROR',
+#             'propagate': True,
+#         },
+#         'SolrClient':  {
+#             'handlers': ['console'],
+#             'level': 'WARN'
+#         },
+# 
+#     }
+# }
```
kmcelwee, line 119: This throws errors if not configured properly and is not directly useful in a local environment, correct?

README.rst
rlskoeser: I see! The instructions don't specify where to run this command, and you - quite logically! - ran it inside the git checkout directory. That would definitely be problematic. I'm not sure it's actually a good idea to have the virtualenv inside the project directory, though - it seems like it could cause problems.  @thatbudakguy where do you put yours? 

`venv` might not be great because that's the name of the module in python3; I could see using `env`, though if we want something generic. 

I also think we should also update this to use a different command, so we don't require installing `virtualenv`:
`python3 -m venv ppa -p python3.6`

README.rst
```diff
@@ -90,6 +90,8 @@ before running the server.
 
 - To build all styles and js for production, including semantic UI::
 
+    npm install
+    npm install -D
```
rlskoeser, line 94: Why is the `install -D` needed? 

We might want to adjust the language a little here. Funny we missed including the installation step!

README.rst
```diff
@@ -189,6 +191,17 @@ In the ``sphinx-docs`` folder, use ``make docs`` to build the HTML documents
...
+-------------
+
+Trouble launching Solr? Try configuring Java to version 8. Later versions of 
+Java will cause Solr 6 to time out.
+
+Solr changes not reflected in search results? ``solrconfig.xml`` must be 
+updated in Solr's main directory: ``solr/server/solr/[CORE]/conf/solrconfig.xml``
+
+We currently do not support Python 3.7 or Solr 7 and above.
+
```
rlskoeser, line 204: These are good things to document, but I'm not sure they belong in the readme. Let's create a DEVELOPERNOTES.rst document with this info.

ppa/local_settings.py.sample
rlskoeser: It probably isn't much of a security concern, but it's not great, and some of the automated code tools we use will flag it and complain. Let's leave it blank, but add a note in the installation steps to explicitly state that you should set a secret key. 

Are there any other local settings we don't mention in the installation instructions that we should call out that need editing or review?

ppa/local_settings.py.sample
rlskoeser: The admin email handler isn't a logging config we'd want in a dev environment, but logging in general is quite useful and it would be nice if the local settings included a good default dev configuration.  Would you update the sample config with this one from mine?
```
LOGGING = {
    'version': 1,
    'disable_existing_loggers': True,
    'formatters': {
        'basic': {
            'format': '[%(asctime)s] %(levelname)s:%(name)s::%(message)s',
            'datefmt': '%d/%b/%Y %H:%M:%S',
        },
    },
    'handlers': {
        'console': {
            'class': 'logging.StreamHandler',
            'formatter': 'basic',
            'level': 'DEBUG'

        },
    },
    'loggers': {
        'django': {
            'handlers': ['console'],
            'level': os.getenv('DJANGO_LOG_LEVEL', 'WARN'),
        },
        'ppa':  {
            'handlers': ['console'],
            'level': 'DEBUG'
        },
        'SolrClient':  {
            'handlers': ['console'],
            'level': 'WARN'
        },
    }
}
```

requirements.txt
```diff
@@ -20,6 +20,6 @@ django-csp
 django-webpack-loader
 requests_oauthlib
 wand
-beautifulsoup4
+beautifulsoup4<4.6.1
```
rlskoeser, line 23: Could you add a comment noting why this is pinned? It's harder to update dependencies later on if we don't know why we pinned something — IIRC this was a wagtail issue, which might not be obvious.

test-requirements.txt
```diff
@@ -1,4 +1,4 @@
-pytest>=3.6
+pytest>=3.6,<5.4
```
rlskoeser, line 1: What errors did you get with pytest 5.4? I just tried locally with pytest 5.4.1 and it worked for me.

Asking because I'd prefer not to pin it if we don't have to, and if we do I'd like to know why so I have some sense of how difficult it will be to update.

README.rst
rlskoeser: Another thought would be to pull all of the installation and setup info into a separate file from the readme, and then a troubleshooting section would be perfectly appropriate. It seems like it might already be long enough to warrant that, what do you think?

README.rst
```diff
@@ -62,8 +62,8 @@ Initial setup and installation:
   solr config, using the core/collection names for development and
   testing that you configured in local settings::
 
-    solr create -c SOLR_CORE -d solr_conf
-    solr create -c SOLR_TEST_CORE -d solr_conf
+    solr create -c ppa -d solr_conf
+    solr create -c ppa-test -d solr_conf
```
rlskoeser, line 66: Reasonable! Then this will match exactly what's in the sample local settings, right? That's good.

README.rst
thatbudakguy: I always put mine in the project directory, fwiw. I also always name them `venv` so that my alias for `source venv/bin/activate` works in any folder.

README.rst
thatbudakguy: it's not needed; this is my miscommunication. npm installs `devDependencies` by default, so `npm install` is sufficient to install everything in `package.json`.

README.rst
thatbudakguy: I'm OK with keeping all of this in a README, but don't have strong opinions. We do list the supported version of python at the top with other versions; maybe we could just add "Java 8" to that list.

README.rst
kmcelwee: Correct, it's one fewer thing to change

test-requirements.txt
kmcelwee:
```
INTERNALERROR>   File "/Users/kevinmcelwee/anaconda3/envs/py36/lib/python3.6/site-packages/pytest_sugar.py", line 176, in pytest_configure
INTERNALERROR>     sugar_reporter = SugarTerminalReporter(standard_reporter)
INTERNALERROR>   File "/Users/kevinmcelwee/anaconda3/envs/py36/lib/python3.6/site-packages/pytest_sugar.py", line 214, in __init__
INTERNALERROR>     self.writer = self._tw
INTERNALERROR> AttributeError: can't set attribute
```

It looks like a Java error that comes from `pytest_sugar`. Going to `pytest < 5.4` fixed it for me. Thoughts?

rlskoeser: Please try uninstaling `pytest-sugar` and checking with the latest pytest. If that works, then remove it from the dev requirements and unpin pytest. We don't need it, it just makes the test output nicer (and I think it started causing problems on travis-ci recently too).

README.rst
kmcelwee: And I ended up using anaconda. We could just say "create and activate a python virtual environment with Python 3.6"

requirements.txt
kmcelwee:
```
ERROR: wagtail 2.8.1 has requirement beautifulsoup4<4.6.1,>=4.5.1, but you'll have beautifulsoup4 4.9.0 which is incompatible.
```

(let me know I should comment this elsewhere)

rlskoeser: Please add a one-line comment in the requirements file next to the beautiful soup requirement noting that it's pinned because of wagtail 2.8.1

README.rst
rlskoeser: 🤔 I think I'm ok with that. That seems better than explicit directions that can cause problems, and probably to set this up you need some experience with python anyway.

README.rst
kmcelwee: I've moved Troubleshooting to a DEVELOPERNOTES.rst document and I've added Java and Solr versioning to the top of README.rst

rlskoeser: Agree - python version is documented at the top of the readme; please add solr version and note the java version in reference to Solr. 

While you're updating that line, would you correct the Django version? We're on 2.2 in develop now but I didn't update that yet.

test-requirements.txt
kmcelwee: perfect, it's all good

requirements.txt
kmcelwee: Oh that kind of comment, got it

rlskoeser: Yes, when I reread I saw my request was unclear! You did add a comment like I asked 😁

Issue #330, opened 2020-04-21
add punctuation search


kmcelwee:
We currently use KeywordTokenizerFactory and StandardTokenizerFactory, which strip punctuation from search. SimplifiedRegularExpressionPatternTokenizer or WhitespaceTokenizerFactory can be added to `ppa/solr.py` to allow for punctuation search.

RSK (via #274):
For the unstemmed version, we should consider using a very simple regex tokenization (maybe [Simplified Regular Expression Pattern Tokenizer](https://lucene.apache.org/solr/guide/6_6/tokenizers.html#Tokenizers-SimplifiedRegularExpressionPatternTokenizer) that doesn't throw away all punctuation (which the default tokenization does), because it would be valuable to search on punctuation like / or | that is used to mark meter. This version of the index may also better support searching for things like AB rhyme scheme notation (since A normally gets thrown out as a stop word).

kmcelwee:
Whitespace Tokenizer was implemented, but appropriately configuring the view didn't work out. Might need more complicated solr query. Many results are empty. Changes are found in `solr-stem-punc`

kmcelwee:
Recap for clarification: while #335 was merged into master, the WhitespaceTokenizer was commented out: https://github.com/Princeton-CDH/ppa-django/blob/master/ppa/archive/solr.py#L94

PR #335, opened 2020-04-28
Feature/boosting and punc

kmcelwee:
* increased boosting from 50 to 100 on non-stemmed content for #274
* included whitespace tokenizer that allows for punctuation to be included in query for #330

### DIFF
--- a/ppa/archive/solr.py
+++ b/ppa/archive/solr.py
@@ -91,6 +91,27 @@ class SolrSchema(object):
                 ]
             },
         },
+        {
+            'name': 'punctuation_parser',
+            "class":"solr.TextField",
+            # for now, configuring index and query analyzers the same
+            # if we want synonyms, query must be separate
+            "analyzer": {
+                # "charFilters": [],
+                "tokenizer": {
+                    "class": "solr.WhitespaceTokenizerFactory",
+                },
+                "filters": [
+                    {"class": "solr.StopFilterFactory", "ignoreCase": True,
+                     "words": "lang/stopwords_en.txt"},
+                    {"class": "solr.LowerCaseFilterFactory"},
+                    # {"class": "solr.EnglishPossessiveFilterFactory"},
+                    # {"class": "solr.KeywordMarkerFilterFactory"},
+                    # {"class": "solr.PorterStemFilterFactory"},
+                    {"class": "solr.ICUFoldingFilterFactory"},
+                ]
+            }
+        },
     ]
 
     #: solr schema field definitions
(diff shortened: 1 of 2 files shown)


ppa/archive/solr.py
```diff
@@ -91,6 +91,27 @@ class SolrSchema(object):
...
+            "analyzer": {
+                # "charFilters": [],
+                "tokenizer": {
+                    "class": "solr.WhitespaceTokenizerFactory",
+                },
+                "filters": [
+                    {"class": "solr.StopFilterFactory", "ignoreCase": True,
+                     "words": "lang/stopwords_en.txt"},
+                    {"class": "solr.LowerCaseFilterFactory"},
+                    # {"class": "solr.EnglishPossessiveFilterFactory"},
```
kmcelwee, line 108: Did not include possessive, stemming, or keyword-ing to the punctuation tokenizer. For stemming and keyword-ing, I was afraid "loving," would get stemmed to "love". And possessive would remove apostrophe.

ppa/archive/solr.py
```diff
@@ -91,6 +91,27 @@ class SolrSchema(object):
                 ]
             },
         },
+        {
+            'name': 'punctuation_parser',
```
rlskoeser, line 95: "parser" doesn't make sense to me as a field name, maybe something like `text_punctuation` or `text_with_punctuation`? Strictly speaking, this isn't specific to punctuation per se, that's just one of our use cases. What about something like `text_simpletoken`

ppa/archive/solr.py
```diff
@@ -91,6 +91,27 @@ class SolrSchema(object):
...
+            "class":"solr.TextField",
+            # for now, configuring index and query analyzers the same
+            # if we want synonyms, query must be separate
+            "analyzer": {
+                # "charFilters": [],
+                "tokenizer": {
+                    "class": "solr.WhitespaceTokenizerFactory",
+                },
+                "filters": [
+                    {"class": "solr.StopFilterFactory", "ignoreCase": True,
```
rlskoeser, line 105: I think we should omit the stop word filter factory - I'm hoping this will let us search on A/B rhyme scheme notation, so we don't want to throw out all the "a"s.

solr_conf/solrconfig.xml
```diff
@@ -764,8 +764,9 @@
           pub_place
           publisher
           source_id^10
-          content_nostem^50
+          content_nostem^100
           content
+          content_punctuation^100
```
rlskoeser, line 769: I think the punctuation field should have _no_ boosting at all. It seems likely that there's going to be a lot of junk in this field that will give us extraneous matches, so we don't want it boosted.

solr_conf/solrconfig.xml
```diff
@@ -774,8 +775,9 @@
           subtitle^35
           author^100
           notes^50
-          content_nostem^50
+          content_nostem^100
           content
```
rlskoeser, line 779: Actually, I think we also should boost the regular content so that it will have priority over matches in the simple tokenized content. That probably means you need to boost the nostem content even more.

PR #346, opened 2020-09-03
Refactor to use parasolr and remove dependency on SolrClient

rlskoeser:
Updates ppa to use parasolr and refactors the join query to explicitly use edismax query syntax for the keyword search and use grouping for edismax and join queries to address a bug in the archive search.

I haven't written tests against the new ArchiveSearchQuerySet directly, but it is tested very thoroughly through the search view tests so I think that might be enough at this point.

I noticed there are still a couple of TODOs that got flagged, I probably should address them before we merge, but it would be helpful to get some feedback on all the work I've done so far.

--- 
additional unit tests
- [x] test page indexable in archive models
- [ ] ArchiveSearchQuerySet in solr (currently has full test coverage; tested indirectly through the view, but better to test directly)


ci/testsettings.py
```diff
@@ -28,16 +28,15 @@
...
-        'COLLECTION': 'test-ppa',
-        'URL': 'http://localhost:8983/solr/',
-        'ADMIN_URL': 'http://localhost:8983/solr/admin/cores'
-    },
     'default': {
-        'COLLECTION': 'ppa-pa11y',
         'URL': 'http://localhost:8983/solr/',
-        'ADMIN_URL': 'http://localhost:8983/solr/admin/cores'
-    },
+        'COLLECTION': 'ppa-pa11y',
```
thatbudakguy, line 33: we should probably just call the default core `ppa`; I imagine `pa11y` will get removed before long.

ppa/archive/models.py
```diff
@@ -823,3 +782,70 @@ def get_hathi_data(self):
...
+    parasolr index manage command.'''
+
+    @classmethod
+    def items_to_index(cls):
+        '''Return a generator of page data to be indexed, with data for
+        pages for all works returned by  :meth:`DigitizedWork.items_to_index`
+        '''
+        for work in DigitizedWork.items_to_index():
+            for page_data in Page.page_index_data(work):
+                yield page_data
```
thatbudakguy, line 798: this is so rad, I love the way the method looks - easy to read and understand

ppa/archive/views.py
```diff
@@ -144,129 +131,50 @@ def get_queryset(self, **kwargs):
...
+        solr_pageq = SolrQuerySet().search(content='(%s)' % self.query) \
+            .search(id__in=page_ids) \
+            .only('id') \
+            .highlight('content', snippets=3, method='unified')
+        # populate the result cache with number of rows specified
+        solr_pageq.get_results(rows=len(page_ids))
+        # print("### %d page ids" % len(page_ids))
+        # solr_pageq.get_results(rows=150)
+
+        # FIXME max rows needed?
```
thatbudakguy, line 177: some leftover comments here

ppa/archive/views.py
```diff
@@ -283,27 +191,26 @@ def get_context_data(self, **kwargs):
...
-            else:
-                # NOTE: this error should possibly be raised; 500 error?
-                # or set status on the response?
-                error_msg = 'Something went wrong.'
-            context['error'] = error_msg
+            # if 'Cannot parse' in str(solr_err):
+            #     error_msg = 'Unable to parse search query; please revise and try again.'
+            # else:
+            # NOTE: this error should possibly be raised; 500 error?
+            # or set status on the response?
```
thatbudakguy, line 212: more leftover comments

templates/snippets/footer.html
```diff
@@ -34,7 +34,6 @@
         </section>
         <div class="contact ui inverted horizontal link list">
             <span class="item">©2018 Trustees of Princeton University</span>
-            <a class="item" href="https://accessibility.princeton.edu/">Accessibility</a>
```
thatbudakguy, line 37: why remove this?

thatbudakguy (approved):
everything looks great; just saw some leftover comments which I imagine you're going to clean up anyway. nice work!

ci/testsettings.py
rlskoeser: I wondered about that; wasn't sure if something depended on it but now that I'm checking I can't find any references. I'll rename.

templates/snippets/footer.html
rlskoeser: In the the last review he did for us (on S&co), JJ said we didn't need to link to the accessibility site in our footer, just to the accessibility assistance. Thought I would update here for consistency now that we know (IDK why I did it on the feature branch, though!).

thatbudakguy: oh! good to know

ci/testsettings.py
rlskoeser: Renaming caused an error, and it didn't seem trivial to update, so I just reverted the change. We can deal with it when we switch from pa11y to axe.

PR #349, opened 2020-09-16
Use solr KeywordRepeatFilterFactory instead of unstemmed field copies

rlskoeser:
Simpler solution for unstemmed search — instead of creating duplicate fields with and without stemming, we use Solr's keyword repeat filter to duplicate the unstemmed words before stemming is applied.

See https://lucene.apache.org/solr/guide/6_6/language-analysis.html#LanguageAnalysis-KeywordRepeatFilterFactory


thatbudakguy (approved):
lgtm; if this does what solr says it does then we should be good

PR #350, opened 2020-09-16
Revise admin add from hathi to use rsync instead of data api

rlskoeser:
I think this is basically working, although it could stand to be tested more. Would appreciate any thoughts on better error checking and reporting on the rsync.

### DIFF
--- a/ppa/archive/admin.py
+++ b/ppa/archive/admin.py
@@ -94,7 +94,7 @@ class DigitizedWorkAdmin(admin.ModelAdmin):
             request, form, formsets, change
         )
         digwork = DigitizedWork.objects.get(id=form.instance.pk)
-        digwork.index(params={"commitWithin": 10000})
+        digwork.index()
 
     def bulk_add_collection(self, request, queryset):
         '''
--- a/ppa/archive/util.py
+++ b/ppa/archive/util.py
@@ -96,8 +96,10 @@ class HathiImporter:
         # create temp file with list of paths to synchronize
         with tempfile.NamedTemporaryFile(prefix='ppa_hathi_pathlist-',
                                          suffix='.txt', mode='w+t') as fp:
-            for path in self.rsync_file_paths():
-                fp.write(path)
+            file_paths = list(self.rsync_file_paths())
+            # sorting may make rsync more efficient
+            file_paths.sort()
+            fp.writelines(file_paths)
 
             # flush to make content available to rsync
             fp.flush()
--- a/ppa/archive/util.py
+++ b/ppa/archive/util.py
@@ -16,7 +16,7 @@ from pairtree.pairtree_path import id_to_dirpath
 from parasolr.django.signals import IndexableSignalHandler
 
 from ppa.archive import hathi
-from ppa.archive.models import DigitizedWork
+from ppa.archive.models import DigitizedWork, Page
 
 
 logger = logging.getLogger(__name__)
@@ -110,6 +110,7 @@ class HathiImporter:
                 'src': settings.HATHITRUST_RSYNC_PATH,
                 'dest': settings.HATHI_DATA
             }
+            logger.debug('rsync command: %s' % rsync_cmd)
             try:
                 subprocess.run(args=rsync_cmd.split(), check=True)
                 # NOTE: when we get to python 3.7, use capture_output
@@ -128,10 +129,20 @@ class HathiImporter:
         # disconnect indexing signal handler before adding new content
         IndexableSignalHandler.disconnect()
 
+        # use rsync to copy data from HathiTrust dataset server
+        # to the local pairtree datastore for ids to be imported
+        logger.info('rsyncing pairtree data for %s', ', '.join(self.htids))
+        self.rsync_data()
+        # FIXME: need better error handling here! rsync can error
+        # or timeout; should we capture output and report that?
+        # Indexing logs an error if pairtree is not present for an
+        # unsuppressed work; perhaps we could do a similar check here?
+
         for htid in self.htids:
             try:
+                # fetch metadata and add to the database
                 digwork = DigitizedWork.add_from_hathi(
-                    htid, self.bib_api, get_data=True,
+                    htid, self.bib_api,
                     log_msg_src=log_msg_src, user=user)
                 if digwork:
                     self.imported_works.append(digwork)
@@ -160,7 +171,7 @@ class HathiImporter:
             DigitizedWork.index_items(self.imported_works)
             for work in self.imported_works:
                 # index page index data in chunks (returns a generator)
-                DigitizedWork.index_items(work.page_index_data())
+                DigitizedWork.index_items(Page.page_index_data(work))
 
     def get_status_message(self, status):
         '''Get a readable status message for a given status'''
--- a/ppa/archive/util.py
+++ b/ppa/archive/util.py
@@ -11,6 +11,7 @@ import tempfile
 from collections import OrderedDict
 from json.decoder import JSONDecodeError
 
+from cached_property import cached_property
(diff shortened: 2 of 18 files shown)


ppa/archive/util.py
```diff
@@ -68,6 +77,48 @@ def filter_existing_ids(self):
...
+                'path_file': fp.name,
+                'server': settings.HATHITRUST_RSYNC_SERVER,
+                'src': settings.HATHITRUST_RSYNC_PATH,
+                'dest': settings.HATHI_DATA
+            }
+            logger.debug('rsync command: %s' % rsync_cmd)
+            try:
+                subprocess.run(args=rsync_cmd.split(), check=True)
+                # NOTE: when we get to python 3.7, use capture_output
+                # to report on what was done or any errors
```
kmcelwee, line 117: Oh 😢 sorry to hear that this feature isn't available with our current version of python. But `subprocess.run` still returns an object with attribute `returncode` where we can get an idea of what kind of error rsync ran, correct?

kmcelwee: Rsync error codes:
```
0     Success
 1     Syntax or usage error
 2     Protocol incompatibility
 3     Errors selecting input/output files, dirs
 4     Requested action not supported: an attempt was made to manipulate 64-bit
       files on a platform that cannot support them; or an option was specified
       that is supported by the client and not by the server.
 5     Error starting client-server protocol
 6     Daemon unable to append to log-file
10     Error in socket I/O
11     Error in file I/O
12     Error in rsync protocol data stream
13     Errors with program diagnostics
14     Error in IPC code
20     Received SIGUSR1 or SIGINT
21     Some error returned by waitpid()
22     Error allocating core memory buffers
23     Partial transfer due to error
24     Partial transfer due to vanished source files
25     The --max-delete limit stopped deletions
30     Timeout in data send/receive
35     Timeout waiting for daemon connection
```
[from stack overflow](https://stackoverflow.com/questions/20737204/comprehensive-list-of-rsync-error-codes)

ppa/archive/util.py
```diff
@@ -68,6 +77,48 @@ def filter_existing_ids(self):
...
+            except subprocess.CalledProcessError:
+                logger.error('HathiTrust rsync failed (command %s)' %
+                             rsync_cmd)
```
kmcelwee, line 120: something along these lines might provide more information:
```python
except subprocess.CalledProcessError as err:
     logger.error(err)
```
outputs:
```
Command '['rsync', ...args]' returned non-zero exit status #.
```

kmcelwee (commented):
It looks like rsync removed a lot of code, which is exciting!

Random sidenote from rsync that we might want to consider if things get slow:
```
NOTE:  sorting the list of files in the --files-from input helps
              rsync to be more efficient, as it  will  avoid  re-visiting  the
              path  elements that are shared between adjacent entries.
```

ppa/archive/util.py
```diff
@@ -68,6 +77,48 @@ def filter_existing_ids(self):
...
+    # NOTE: add -v if needed for debugging
+    rsync_cmd = 'rsync -rLt --delete --ignore-errors ' \
+                + ' --files-from=%(path_file)s %(server)s:%(src)s %(dest)s'
+
+    def rsync_data(self):
+        # create temp file with list of paths to synchronize
+        with tempfile.NamedTemporaryFile(prefix='ppa_hathi_pathlist-',
+                                         suffix='.txt', mode='w+t') as fp:
+            for path in self.rsync_file_paths():
+                fp.write(path)
```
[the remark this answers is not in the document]
rlskoeser: Oh, good call — yes, each path should be on a separate line. (I think maybe I only tested with one? 🙄 )

And I bet I can just use `writelines` here instead of looping.

ppa/archive/util.py
rlskoeser: Return code would tell me if the whole thing failed — which I should definitely add checking for, e.g. if the connection fails — thanks for this list of error codes! Helpful. 

But I don't think this is granular enough — I *think* it could return without an error but not transfer all of the files we expect. I guess I could test that, though... 🤔

ppa/settings.py
```diff
@@ -252,6 +252,11 @@
     },
 }
 
+
+# defaults for HathiTrust dataset rsync, per their documentation
+HATHITRUST_RSYNC_SERVER = 'datasets.hathitrust.org'
+HATHITRUST_RSYNC_PATH = ':ht_text_pd'
```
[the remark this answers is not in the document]
rlskoeser: A single colon is the normal separator between the server name and the path on the server, like you would use with scp. You use two colons when you want to specify an rsync module instead of a path — which seems to be how they have it configured.

I put the colon here in the config to make it optional; that way, we can override it and use a regular file path on our qa server in local settings for testing without having to configure an rsync module on the server.

rlskoeser:
@kmcelwee thanks for the review. Good to know about sorting the files, easy enough to do that and happy to have anything that will make it more efficient!

ppa/archive/admin.py
```diff
@@ -94,7 +94,7 @@ def save_related(self, request, form, formsets, change):
             request, form, formsets, change
         )
         digwork = DigitizedWork.objects.get(id=form.instance.pk)
-        digwork.index(params={"commitWithin": 10000})
+        digwork.index()
```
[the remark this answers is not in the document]
rlskoeser: @kmcelwee sorry, missed responding to this. Removing it because we dropped that argument when we adapted original solr code in PPA for parasolr. 

I don't follow your comment about rsync operating across the entire database, though.

rlskoeser:
@kmcelwee thanks for providing the rsync error codes, I don't know that I would have gone to the trouble to look them up myself, but since you provided them I incorporated them into the log error if we get a non-zero result.

Updated to check if the rsync succeeds before importing, and added an rsync error status to the list of possible results.

ppa/archive/util.py
```diff
@@ -139,12 +174,22 @@ def add_items(self, log_msg_src=None, user=None):
         # unsuppressed work; perhaps we could do a similar check here?
 
         for htid in self.htids:
+            # if rsync did not create the expected directory,
+            # set error code and bail out
+            expected_path = os.path.join(settings.HATHI_DATA,
+                                         self.pairtree_paths[htid])
+            if not os.path.isdir(expected_path):
+                self.results[htid] = self.RSYNC_ERROR
+                continue
```
kmcelwee, line 183: Right, I hadn't realized that you wanted to check that each new ID had properly synced. This looks good 🙂

ppa/archive/util.py
```diff
@@ -32,13 +33,17 @@ class HathiImporter:
...
     SKIPPED = 2
+    #: rsync error
+    RSYNC_ERROR = 3
 
     #: human-readable message to display for result status
     status_message = {
         SUCCESS: 'Success',
         SKIPPED: 'Skipped; already in the database',
         hathi.HathiItemNotFound: 'Error loading record; check that id is valid.',
+        # possibly irrelevant with removal of data api code
```
kmcelwee, line 44: It's referenced mostly through `TestHathiBaseAPI` in [test_hathi.py](https://github.com/Princeton-CDH/ppa-django/blob/c8c66d620129fa0dab6301661ba33450a53c750d/ppa/archive/tests/test_hathi.py#L173) Am I right in guessing this PR makes that class unnecessary as well?

kmcelwee (approved):
Looks good!

ppa/archive/util.py
rlskoeser: We still need Hathi API code for the bibliographic API. The comments indicate the 403 is specific to data api, but it seems like it could still occur, so I'm going to leave the exception and error handling in for now.

PR #352, opened 2020-09-21
chore/page indexing multiproc

rlskoeser:
Custom page indexing script with multiprocessing.

Haven't written any unit tests yet because I wanted to get feedback first. 

Would love to get reports on index times & cpu count on your machines; also interested in comparison with index time before, if you remember roughly what it was.

### DIFF
--- a/ppa/archive/management/commands/index_pages.py
+++ b/ppa/archive/management/commands/index_pages.py
@@ -105,5 +105,5 @@ class Command(BaseCommand):
             for item_type, total in facets.facet_fields.item_type.items():
                 item_totals.append('%d %s%s' % (
                     total, item_type, '' if total == 1 else 's'))
-            print('\nItems in Solr by item type: %s' %
-                  (', '.join(item_totals)))
+            self.stdout.write('\nItems in Solr by item type: %s' %
+                              (', '.join(item_totals)))
--- a/ppa/archive/management/commands/index_pages.py
+++ b/ppa/archive/management/commands/index_pages.py
@@ -73,7 +73,7 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            '--cpus', default=cpu_count(),
+            '--cpus', default=cpu_count(), type=int,
             help='Number of CPUs to use (cpu_count by default: %(default)s)')
 
     def handle(self, *args, **kwargs):
--- a/ppa/archive/management/commands/index_pages.py
+++ b/ppa/archive/management/commands/index_pages.py
@@ -71,6 +71,10 @@ class Command(BaseCommand):
     '''Index page data in Solr (multiprocessor implementation)'''
     help = __doc__
 
+    #: normal verbosity level
+    v_normal = 1
+    verbosity = v_normal
+
     def add_arguments(self, parser):
         parser.add_argument(
             '-p', '--processes', default=cpu_count(), type=int,
@@ -78,9 +82,12 @@ class Command(BaseCommand):
                  '(cpu_count by default: %(default)s)')
 
     def handle(self, *args, **kwargs):
+        self.verbosity = kwargs.get('verbosity', self.v_normal)
+        if self.verbosity >= self.v_normal:
+            self.stdout.write('Indexing with %d processes' %
+                              max(2, kwargs['processes']))
         work_q = Queue()
         page_data_q = Queue()
-
         # populate the work queue with digitized works that have
         # page content to be indexed
         for digwork in DigitizedWork.items_to_index():
--- a/ppa/archive/management/commands/index_pages.py
+++ b/ppa/archive/management/commands/index_pages.py
@@ -19,8 +19,8 @@ def page_index_data(work_q, page_data_q):
     for and a queue where page data will be added.'''
     while True:
         try:
-            digwork = work_q.get(True, 1)
             # convert the generator to a list
+            digwork = work_q.get(timeout=1)
             # — might be nice to chunk, but most books are small
             # enough it doesn't matter that much
             page_data_q.put(list(Page.page_index_data(digwork)))
@@ -53,16 +53,6 @@ def process_index_queue(index_data_q, total_to_index, work_q):
             # indexer has just gotten ahead of page index data
             if work_q.empty():
                 progbar.finish()
-
-                # print a summary of solr totals by item type
-                facets = SolrQuerySet().all().facet('item_type').get_facets()
-                item_totals = []
-                for item_type, total in facets.facet_fields.item_type.items():
-                    item_totals.append('%d %s%s' % (
-                        total, item_type, '' if total == 1 else 's'))
-                print('\nItems in Solr by item type: %s' %
-                      (', '.join(item_totals)))
-
                 # finish indexing process
                 return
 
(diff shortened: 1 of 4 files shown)


thatbudakguy (approved):
I hit one strange issue while testing this - evidently macOS versions past High Sierra have a security feature that interacts with python's multiprocessing module. I saw this output when running the script:
```
objc[67570]: +[__NSPlaceholderDate initialize] may have been in progress in another thread when fork() was called.
objc[67570]: +[__NSPlaceholderDate initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```
The script appeared to run (fans got loud), but there was no other output and the progress bar didn't render. The solution is to set an environment variable in order to allow multiprocessing to bypass a security restriction, as described in [this stackoverflow answer](https://stackoverflow.com/a/52230415). After that, the script ran normally and I could see the progressbar.

In terms of time saved, the predicted total time is hovering around 1hr for me. Previously it had been about 3hrs, so definitely a substantial speedup in any case. I used the detected default of 4 processes on my machine.

Only suggestion re: implementation is maybe add a line in the script to log how many processes are being used, since it's nice to know what's actually going on if you didn't specify anything to the command (something like `indexing using 4 processes`).

rlskoeser:
@thatbudakguy good to know about the security setting; any thoughts on where to document that? 

I like your suggestion about reporting the number of processes that will be used, will add that.

thatbudakguy:
good question - could be tricky because idk if people's instinct would be to open the actual manage command file or anything like that. what do you think about actually checking the environment for `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` and warning if it's not set?

thatbudakguy:
one random other thing that occurred to me - for DIRECT, I saw a small speedup when sorting the texts to index to by length and starting with the longest. obviously the underlying indexing implementation is different (solr), but I think solr internally uses tries, so it might not be _that_ different. if it's easy, you could try sorting the works by their page count and see if it saves any time.

rlskoeser:
> one random other thing that occurred to me - for DIRECT, I saw a small speedup when sorting the texts to index by length and starting with the longest. obviously the underlying indexing implementation is different (solr), but I think solr internally uses tries, so it might not be _that_ different. if it's easy, you could try sorting the works by their page count and see if it saves any time.

Interesting that you suggest that! I did actually try exactly this — sort the works to index longest page count. It didn't actually seem that it made a difference to the overall time, and it made the initial completion estimates really pessimistic, so I scrapped it. I feel like there might be an optimal order, and wondered about sorting and then combining the beginning of the list with the end so that we process the longest and shortest works together ... but based on my test it didn't seem like it was worth more effort.

I'm not sure what you mean about indexing with solr & tries, but also not sure it's relevant — the bottleneck for us is definitely in generating the page index data. If I want to optimize more, I was thinking I should profile the code that generates page index data for the longest volumes and see if there's something I can improve. You may have seen in the code, I had to add logic so the indexer doesn't quit early because it can get ahead of the processes adding page data to the queue — this happens even on my machine running 7 page data processes and 1 indexer!

thatbudakguy:
Makes sense - especially if you already tried and there wasn't much of a change. For me on DIRECT it was a pretty small difference (on the order of like 5%, and obviously haven't tested on any other machine). The reason I thought it might be applicable is because both DIRECT's index (which is basically just a fancy dict with a bloom filter, using spaCy's `Lookups`) and solr's indices use underlying data structures that are sensitive to the amount of data already in them, so sometimes it matters what order you do things in. But it's probably a lost cause trying to deterministically predict solr's indexing performance since it's so complex, and anyway our bottleneck is in getting the data to index, like you said.

rlskoeser:
> what do you think about actually checking the environment for `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` and warning if it's not set? (is it set on your machine, btw?)

It's not set on my machine because I haven't upgraded yet... 🙄   If I had run into it I would have warned you at least in the comments on the PR.

I hesitate to write more code for this, would rather just make a note somewhere since it likely only affects people on this team. I'll add a note to the developer notes that Kevin started.

ppa/archive/management/commands/index_pages.py
```diff
@@ -0,0 +1,106 @@
...
+import progressbar
+
+from ppa.archive.models import DigitizedWork, Page
+
+
+def page_index_data(work_q, page_data_q):
+    '''Function to generate page index data and add it to
+    a queue. Takes a queue with digitized works to generate pages
+    for and a queue where page data will be added.'''
+    while True:
```
kmcelwee, line 20:
```suggestion
    while not work_q.empty():
```
I assume that there was some logic behind why we shouldn't do this? I have an irrational fear of `while True`s

ppa/archive/management/commands/index_pages.py
```diff
@@ -0,0 +1,106 @@
...
+        try:
+            digwork = work_q.get(True, 1)
+            # convert to the generator to a list
```
kmcelwee, line 23:
```suggestion
            # convert the generator to a list
```
Do we really care about nitpicking stuff like this?

ppa/archive/management/commands/index_pages.py
```diff
@@ -0,0 +1,106 @@
...
+                progbar.finish()
+
+                # print a summary of solr totals by item type
+                facets = SolrQuerySet().all().facet('item_type').get_facets()
+                item_totals = []
+                for item_type, total in facets.facet_fields.item_type.items():
+                    item_totals.append('%d %s%s' % (
+                        total, item_type, '' if total == 1 else 's'))
+                print('\nItems in Solr by item type: %s' %
+                      (', '.join(item_totals)))
```
kmcelwee, line 64: Perhaps we could take this out of this function? I feel like we might want to remove any code from Process that doesn't explicitly need multiprocessing.

kmcelwee (commented):
Very cool stuff!

ppa/archive/management/commands/index_pages.py
rlskoeser: @kmcelwee That would make more sense! I tried to put it in the main handle method but couldn't get it to not run before the other work had completed, even though I ran the queue join method first. I tried a few combinations of joins to wait on the process / queue completion and couldn't get it to wait, so I just put it here. (Kind of cheating, I know).

ppa/archive/management/commands/index_pages.py
rlskoeser: I appreciate the edits and corrections if you spot them — anything that makes it more readable later is always better.

ppa/archive/management/commands/index_pages.py
rlskoeser: This would mean calling the empty check on the queue every iteration, which is extra processing that we don't actually need. Also, apparently empty is not completely reliable "because of multithreading/multiprocessing semantics" ?! 🤨

ppa/archive/management/commands/index_pages.py
rlskoeser: I got it working! I think before I was trying to join on the queue (using a `JoinableQueue`); maybe that didn't work because the page data queue can be empty temporarily when a large book is being processed. Joining on the indexer process now and it seems to be working.

PR #362, opened 2021-05-17
Switch to fomantic-ui and update js deps (#357)

thatbudakguy:


rlskoeser (commented):
Can you remind me we why we have a local copy of semantic?

Would you add developer notes about how to update the local copy of semantic/fomantic, or is it handled as part of the npm install/build?

thatbudakguy:
Fomantic actually has [a section in their docs about the advantages of building locally](https://fomantic-ui.com/introduction/build-tools.html#why-build-tools). Most of these reasons are relevant to us, particularly the ones about building only the UI components we use (which you can see in the `semantic.json` config file), using autoprefixer, and theming. We alter "theme variables" to set things like colors and fonts in their LESS sources, so that our SASS has to do as little work as possible "on top".

You can update the fomantic-ui sources like a regular npm package using `npm update` or `npm install`, and you can rebuild it using the `npm run build:semantic` command, which is documented in the README.

rlskoeser:
> Fomantic actually has [a section in their docs about the advantages of building locally](https://fomantic-ui.com/introduction/build-tools.html#why-build-tools). Most of these reasons are relevant to us, particularly the ones about building only the UI components we use (which you can see in the `semantic.json` config file), using autoprefixer, and theming. We alter "theme variables" to set things like colors and fonts in their LESS sources, so that our SASS has to do as little work as possible "on top".
> 
> You can update the fomantic-ui sources like a regular npm package using `npm update` or `npm install`, and you can rebuild it using the `npm run build:semantic` command, which is documented in the README.

Perfect — I hoped that was the case but wanted to confirm (and make sure I understand!). I remembered we did some theming/customization but had forgotten these other aspects.

Issue #338, opened 2020-05-20
move ci to github actions


thatbudakguy:
as noted in https://github.com/Princeton-CDH/pemm-scripts/pull/45#issue-420883706, we can use travis's build matrix (or similar functionality on github actions) to do the python and js parts of the build separately. see the link for more context.

PR #363, opened 2021-05-17
Switch CI to GitHub Actions (#338)

thatbudakguy:
couple of things to note here.

there was an error I didn't understand that looked like:
```py
            for chunk in chunks:
>               asset = assets['assets'][chunk]
E               KeyError: 'assets'
```
which was happening during unit tests; it seemed to be related to attempting to load the static files in tests because the test client was accessing pages over HTTP. specifically, i think `django-webpack-loader` couldn't find the right files, but i verified they were in the right place on the CI runner. ultimately I settled on doing something different from s&co: loading a fake `webpack-stats.json` (now in the `ci/` folder) so that requests for styles/scripts in tests would not error, but instead do nothing. it's based on [this github issue](https://github.com/django-webpack/django-webpack-loader/issues/187).

this approach has some advantages and drawbacks: it means we don't need node, npm, or any static build step at all to run the python unit tests, which makes CI overall faster and simpler. however, you do need to specify the bundles that would be built in `ci/webpack-stats.json` (you can see they're set to an empty array in the JSON; compare a `webpack-stats.json` generated by `npm run build:prod` to see the difference). if we changed our webpack setup by introducing a new bundle, we'd need to update that file.

the fake stats file is loaded by a new line added to `ci/testsettings.py` which references it specifically. to get this to work, I had to make sure that the webpack config was declared prior to loading the localsettings file, which it previously wasn't (presumably so it could depend on the `DEBUG` setting).

i haven't set up anything to prep for the migration to postgres, but i could add a quick matrix build strategy to the python unit tests that only utilizes one matrix cell (solr 6.6 + mariadb) so that we could easily add postgres as an option later.

we could also think about adding lighthouse to this workflow; i think the ease of doing that mostly depends on how good of a test site fixture we have – if there's content that's easy to load for lighthouse to crawl, shouldn't be too difficult.


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,82 @@
...
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+      - run: pip install -r dev-requirements.txt
+      - run: cp ci/testsettings.py ppa/local_settings.py
+      - run: python -c "import uuid; print('SECRET_KEY = \'%s\'' % uuid.uuid4())" >> ppa/local_settings.py
+      - run: py.test --cov=./ --cov-report=xml
+      - uses: codecov/codecov-action@v1
+        with:
+          flags: python
```
rlskoeser, line 68: oh, cool — hadn't seen these before

rlskoeser (approved):
Everything here seems reasonable.

The setup you describe for the webpack file sounds ok as long as it's documented — could you add instructions to the readme or developer notes about when and how to rebuild it?

Please also update the ci badge in the readme to swap out travis for the github actions build

.github/workflows/unit-tests.yml
thatbudakguy: yeah, I _think_ this is working as intended? I checked codecov and it does show both the typescript and python as part of the same coverage report, so fingers crossed it'll continue to work without issue. we used the same setup on s&co, fwiw

PR #364, opened 2021-05-19
hover/focus state changes

thatbudakguy:
this PR:

- adds a dashed black focus indicator for buttons
- adds a grey triangle as an indicator of hover/focus states for each of the main nav items
- adds the same indicator to the items in the "about" submenu (but not the "about" item itself)
- ensures that the "about" item isn't independently focusable

I applied the triangle to the bottom of the PPA logotype when it is focused; lmk if that isn't a reasonable interpretation of the style. See attached video where I hover each of the items, and then focus through each of them using the <kbd>tab</kbd> key.

https://user-images.githubusercontent.com/4924494/118861612-b248d200-b8aa-11eb-8537-7bc2cf865b61.mov

Note also that the order that you <kbd>tab</kbd> focus through the collections buttons is not the order they appear! This is unavoidable; we used special CSS to rearrange the buttons dynamically when they are selected/unselected but that styling doesn't correlate to DOM order. We don't want to rearrange the actual elements in the DOM since that would require javascript. See attached video:

https://user-images.githubusercontent.com/4924494/118862263-82e69500-b8ab-11eb-94c1-cebab9d6785e.mov

### DIFF
--- a/sitemedia/semantic/src/site/elements/button.overrides
+++ b/sitemedia/semantic/src/site/elements/button.overrides
@@ -41,4 +41,9 @@
         color: @disabledTextColor;
         border-color: @disabledTextColor;
     }
+
+    &.focus,
+    &:focus {
+        outline: @focusOutline;
+    }
 }
--- a/sitemedia/semantic/src/site/elements/button.variables
+++ b/sitemedia/semantic/src/site/elements/button.variables
@@ -32,6 +32,7 @@
 /* Focused */
 @focusBackgroundColor: @primaryColor;
 @focusColor: @invertedTextColor;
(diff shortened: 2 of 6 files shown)


rlskoeser (approved):
Everything here looks good from my perspective. Defer to @gissoo on the interpretation of the styles, obviously! Thanks for including the videos to make it easy to review.

gissoo:
@thatbudakguy thank you!! This looks great! I would vote to remove the hover style from the PPA logotype, that element never had hover style anyway and I think it's generally expected that the logotype is clickable and that it can link to the homepage, it doesn't need a hover style, looks a bit odd. Thank you for the videos!!!

PR #382, opened 2021-06-10
Revise label for external link on detail page

rlskoeser:
tweak external link labeling to match what Gale wanted and specified in the MOU 

(existing logic was very close to what we needed already!)


thatbudakguy (approved):
looks good. codefactor's beef seems to be the amount of methods on `DigitizedWork`, but idk if there's an easy refactor that would fix that, and not sure how much it matters.

Issue #377, opened 2021-06-03
As a user, I want to search within a single Gale/ECCO item so that I can find more page results and keywords in context than are available on the main archive search.


rlskoeser:
## dev notes
Should work without additional code changes once import and index is implemented. Adding to explicitly track and test.

mnaydan:
I tested a variety of keywords across CB and CW volumes (and checked a couple HT volumes for good measure). Search within volume functions as expected.

However, in testing this issue I discovered that I was unable to yield expected keyword search results for words appearing on pages that ECCO categorizes under "Front Matter" or "Back Matter" in both the main search and search within volume. For example, "bardism" in CW0115966339 yields zero results (the JSON file didn't show any OCR for pages 2-79, where "bardism" occurs). I tested with CB0130984232, CW0103485860, CB0127060085, CW0113494123 to confirm, and the pattern holds across all. We know Gale has the OCR for the Front/Back Matter because they are searchable in Gale's interface. We suspect this is an issue with the API, and Bill is looking into it. 

@rlskoeser, given that this is most likely an issue on Gale's API end, not your code, should I close this issue? Or do you want me to keep tracking it here?

rlskoeser:
@mnaydan I think we should close this issue, since the problem isn't on our end, and existing code will work once they add the missing ocr content into their api. We might want to track it somehow but I'm not sure where!

mnaydan:
@rlskoeser I think it'll naturally be tracked on the excerpts stories - we won't be able to see text snippets for excerpts if there's no OCR.

PR #383, opened 2021-06-10
Gale/ECCO thumbnails

rlskoeser:
Adjust indexing and page preview template so we can display thumbnails for Gale content

In the process of getting thumbnails working for the search within a single document, I also made the small changes needed for that to work with Gale content #377

### DIFF
--- a/ppa/archive/templates/archive/snippets/page_preview.html
+++ b/ppa/archive/templates/archive/snippets/page_preview.html
@@ -12,6 +12,7 @@ Expected context variables:
     <div class="preview">
         <a href="{% page_url item_id page.order %}"
           target="_blank" rel="noopener noreferrer">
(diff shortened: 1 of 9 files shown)


ppa/archive/templates/archive/snippets/page_preview.html
```diff
@@ -22,6 +23,13 @@
             <img src="{{ 1x_img }}" srcset="{{ 1x_img}}, {{ 2x_img }} 2x"
               alt="page {{ page.label }}"/>
           </noscript>
+          {% elif source == 'Gale' %}
+          {# TODO: move to template tag? combine or separate from hathi? #}
```
thatbudakguy, line 27: agree this makes sense as a template tag, but probably not high priority

thatbudakguy (approved):
looks good!

Issue #378, opened 2021-06-03
As an admin, I want the CSV export to include source so that I can distinguish content from HathiTrust, Gale/ECCO, etc.


rlskoeser:
## testing notes
- [x] use the "download as CSV" link in django admin to download test data, including newly imported content; confirm that there is a "source" column and that it is set accurately to "Gale" or "HathiTrust"

## dev notes
- add source header and source display value to existing admin csv export

mnaydan:
CSV download successful, "Source" column present on CSV with either "HathiTrust" or "Gale" accurately displayed.

PR #384, opened 2021-06-10
Include source in digitized works csv export #378

rlskoeser:


rlskoeser:
this change is so trivial, no need to review

PR #385, opened 2021-06-14
Admin action to bulk-suppress works

rlskoeser:


ppa/archive/admin.py
```diff
@@ -114,6 +116,32 @@ def bulk_add_collection(self, request, queryset):
                                              'to collections')
     bulk_add_collection.allowed_permissions = ('change',)
 
+    def suppress_works(self, request, queryset):
+        # set status to suppressed for every item in the queryset
+        # that is not already suppressed
+        non_suppressed = queryset.exclude(status=DigitizedWork.SUPPRESSED)
+        # save the list of ids being suppressed to update the index after
+        ids_to_suppress = list(non_suppressed.values_list('source_id', flat=True))
+        # change statu in the database
```
thatbudakguy, line 125: "statu" isn't valid english, but it is valid latin!

ppa/archive/admin.py
```diff
@@ -114,6 +116,32 @@ def bulk_add_collection(self, request, queryset):
                                              'to collections')
     bulk_add_collection.allowed_permissions = ('change',)
 
+    def suppress_works(self, request, queryset):
```
thatbudakguy, line 119: i like this naming (`works` to me implies multiple/a queryset) but wonder if it should be consistent with `bulk_add_collection`. should we just rename that one to `add_works_to_collection`? does this even matter?

ppa/archive/admin.py
```diff
@@ -114,6 +116,32 @@ def bulk_add_collection(self, request, queryset):
...
+        # set status to suppressed for every item in the queryset
+        # that is not already suppressed
+        non_suppressed = queryset.exclude(status=DigitizedWork.SUPPRESSED)
+        # save the list of ids being suppressed to update the index after
+        ids_to_suppress = list(non_suppressed.values_list('source_id', flat=True))
+        # change statu in the database
+        updated = non_suppressed.update(status=DigitizedWork.SUPPRESSED)
+        # queryset.update does not trigger save signals;
+        # clear suppressed page + work content from the index
+        solr = SolrClient()
```
thatbudakguy, line 129: idk what goes on behind the scenes here, and it might not be expensive at all, but it seems like you could choose to only initialize the client if there are any ids to suppress (below line 131)

thatbudakguy (approved):
looks good; the check & count for already suppressed items is nice.

ppa/archive/admin.py
rlskoeser: Agree the names are not consistent; I like your suggestion, but also not sure how much it matters...

I'll change if it's easy.

ppa/archive/admin.py
rlskoeser: good point! probably not very expensive, but does require looking up solr connection parameters in django settings, so nice to skip it if we don't need it

PR #386, opened 2021-06-14
Improve Gale API and import script error handling

rlskoeser:
This error handling is for me — I keep forgetting to specify `-c` when I give it a csv file and so it tries to treat the csv as an item, which errors.  

I also decided that any non-ok response should raise an api error rather than return none.

### DIFF
--- a/ppa/archive/management/commands/gale_import.py
+++ b/ppa/archive/management/commands/gale_import.py
@@ -83,7 +83,7 @@ class Command(BaseCommand):
             raise CommandError("A list of IDs or CSV file for is required for import")
 
         # error handling in case user forgets to specify csv file correctly
-        if kwargs.get('id') and len(kwargs['ids']) == 1 and kwargs['ids'][0].endswith('.csv'):
+        if 'ids' in kwargs and len(kwargs['ids']) == 1 and kwargs['ids'][0].endswith('.csv'):
             self.stdout.write(self.style.WARNING(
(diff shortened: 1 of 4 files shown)


ppa/archive/management/commands/gale_import.py
```diff
@@ -82,6 +82,12 @@ def handle(self, *args, **kwargs):
         if not (kwargs["ids"] or kwargs["csv"]):
             raise CommandError("A list of IDs or CSV file for is required for import")
 
+        # error handling in case user forgets to specify csv file correctly
+        if kwargs.get('id') and len(kwargs['ids']) == 1 and kwargs['ids'][0].endswith('.csv'):
+            self.stdout.write(self.style.WARNING(
+                '%s is not a valid id; did you forget to specify -c/--csv?' % kwargs['ids'][0]))
+            return
+
```
thatbudakguy, line 90: is it `id` or `ids` (or both)? first condition in the `if` has `id`. also i sometimes do `if 'id' in kwargs` because i think it reads slightly better, but idk if that's python best practice or anything

thatbudakguy (approved):
the best error handling is the kind that handles errors you _know_ you will make

ppa/archive/management/commands/gale_import.py
rlskoeser: good catch, thank you! pretty sure `if 'id' in kwargs' is more pythonic, will revise

PR #387, opened 2021-06-15
Suppress underline on 'view the archive' button link

rlskoeser:
@thatbudakguy is there a better way to do this? 

there's an `a.archive-link` in _links.scss that specifies no underline, but the `a:not(.ui):not(.card)` style seems to be taking precedence

### DIFF
--- a/srcmedia/scss/pages/_home.scss
+++ b/srcmedia/scss/pages/_home.scss
@@ -27,6 +27,7 @@
 
     .archive-link {
         margin: 0 0 2rem;
+        text-decoration: none;
     }
 
     .asides-collections {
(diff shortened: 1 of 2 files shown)


thatbudakguy (approved):
this solution seems fine, but that selector has caused problems for me before.

links don't have a default underline style (i think because of our css reset), so i wanted to set a low-level one for all links, but then we have a lot of things on this project that are clickable links that don't _look_ like links (cards, etc) which is why the `:not()` is there. not sure why it's taking precedence though; I'd think that stuff like `:not()` adds very little power to a selector (certainly less than a class).

Issue #389, opened 2021-06-16
As an admin, I want a bulk import of MARC metadata from Gale/ECCO so that I can view and search each record by its metadata


mnaydan:
We want to pull the following fields from the MARC records: title, subtitle, sort title, place of publication, publisher, and pub date

mnaydan:
@rlskoeser how is author field formatted in the MARC records? I'm wondering if it's worth pulling author, too, if it's formatted "last name, first name" so that it matches the format of the HT records.

rlskoeser:
@mnaydan yes, planning to pull author from the MARC metadata as well — it should get us a format closer to what we've got already.

rlskoeser:
Here's the output from the most recent run of the gale import script.  (@mnaydan I think I'm still using an old version of your CSV without the most recently corrected IDs).

I'm surprised by how many MARC records we're missing — it didn't seem like that many when I was testing, but when I ran the full import again locally I got similar results. We may need to follow up with Gale to find out what could cause this (could it be a matter of _when_ PUL purchased the marc records?)

I think the character parse errors are because I switched back to using binary MARC records instead of JSON — @mnaydan if you can think of any records with special characters to test, that would be helpful. If anything is getting corrupted we can try switching back to the JSON version of the MARC records.

```
$ ./manage.py gale_import -c /tmp/ECCO_compiled_non-excerpts_fixed_IDs_6.10.21.csv -v 0
MARC record not found for CB0132859609
MARC record not found for CW0112106158
MARC record not found for CW0111720541
MARC record not found for CW0114229753
MARC record not found for CW0114728738
Unable to parse character 0xcc in g0=66 g1=69
MARC record not found for CW0114704032
MARC record not found for CW0110970244
MARC record not found for CW0110970663
MARC record not found for CW0111406390
MARC record not found for CW0109771665
MARC record not found for CW0113390248
MARC record not found for CW0110986375
MARC record not found for CW0110004195
MARC record not found for CW0110433388
MARC record not found for CW0114905944
MARC record not found for CW0115390412
MARC record not found for CW0113836889
MARC record not found for CW0113837303
MARC record not found for CW0114032230
MARC record not found for CW0114032713
MARC record not found for CW0115236604
MARC record not found for CW0115419546
MARC record not found for CW0115420029
MARC record not found for CW0115963418
MARC record not found for CW0119571809
MARC record not found for CW0119572297
MARC record not found for CW0120106050
MARC record not found for CW0120106500
[...]

Issue #379, opened 2021-06-03
As a user, I want to add a Gale/ECCO work to my Zotero library from the item page or the search results page, so that I can save references for later research or citation.


rlskoeser:
## testing notes

Test importing a gale item from list view or from detail view — it should work the same as for HathiTrust materials. Metadata will be coming from the marc records (you can confirm based on the sid in the view url).

The only exception should be the records that don't have marc records in the set we were given.

## dev notes
Builds on #36 and #35

If `DigitizedWork.get_metadata` has already been updated as part of #369, then this will likely work for the current implementation (MARC only) without further effort.

mnaydan:
Tested with Holmes (CW0110145213) on detail view and it imported successfully. URL confirms metadata coming from MARC record:
(https://link.gale.com/apps/doc/CW0110145213/ECCO?sid=gale_marc&u=prin77918). 

Also tested importing multiple items from list view (Bysshe) and it was successful. 

Works without MARC records save as webpage (as expected).

One weird thing is that page count in Zotero doesn't match page count on our site or in ECCO? For example, the Zotero record for Holmes shows 18 pages, whereas [on our site it says 189](https://test-ppa.cdh.princeton.edu/archive/CW0110145213/). This is happening with the Bysshe too, for instance [this record is 523 pages](https://test-ppa.cdh.princeton.edu/archive/CW0115898745/) but in Zotero it says 2. @rlskoeser thoughts?

rlskoeser:
@mnaydan good testing as always, catching the page number discrepancy! I looked at the MARC record for the Bysshe document you linked with 523 pages to check this. If I'm reading the MARC documentation correctly, page numbers would be included in the [extent subfield](https://www.loc.gov/marc/bibliographic/bd300.html) (300 a), but that can include other extent information. For the Bysshe marc record, the contents of that field is "2v", presumably indicating 2 volumes — but I guess Zotero is just taking the numeric portion.  (We could check or ask on the zotero forums if you think this is important to track down.)

[...]

PR #390, opened 2021-06-16
Incorporate Gale/ECCO MARC records for import and zotero metadata #389 #379

rlskoeser:
We were given MARC records in binary format for ECCO1 and ECCO2, but retrieving a particular MARC record for an item by Gale ID was far too slow (~30 seconds on the first one I tried in ECCO2, which is smaller than ECCO1).

My solution is to split those MARC records into a local pairtree datastore keyed on Gale item id, and serialize it as JSON (in my tests, JSON was smaller on the filesystem and faster to parse when loading records). This makes it possible to quickly retrieve the MARC records we need on demand, for import or for metadata harvesting (zotero), based on Gale item id.

Changes included in this PR:
- a new script to split out the marc records and a method to access marc record by id
- split out hathitrust bibdata and marc metadata so the marc portion could be reused for Gale content (code change here is almost entirely cut & paste; only change is publication date handling, to extract from marc for Gale)
- use marc records for metadata in gale import script
- serve out marc record for Gale items (for zotero harvest)

Might need more error handling? But I'm not sure yet what kind of errors are likely & important enough to require code handling.


ppa/archive/management/commands/split_marc.py
```diff
@@ -0,0 +1,58 @@
...
+                    # Gale item id used for API is only available in the
+                    # access url; split the url and get the id
+                    gale_id = record["856"]["u"].split("/")[5]
+                    # create an object in the pairtree
+                    pmarc = marc_store.get_object(gale_id, create_if_doesnt_exist=True)
+                    # use pymarc to export as JSON (more compact & faster than XML)
+                    output = StringIO()
+                    writer = pymarc.JSONWriter(output)
+                    writer.write(record)
+                    writer.close(close_fh=False)  # important for valid json!
```
thatbudakguy, line 45: is it `writer.close()` that's important, or passing `close_fh=False`? i'm kinda unclear how this works – if the file handle isn't closed, would it automatically be closed once this iteration of the forloop completes and `writer` goes out of scope?

ppa/archive/management/commands/split_marc.py
```diff
@@ -0,0 +1,58 @@
...
+                reader = pymarc.MARCReader(marcfile, to_unicode=True)
+                for record in reader:
+                    # Gale item id used for API is only available in the
+                    # access url; split the url and get the id
+                    gale_id = record["856"]["u"].split("/")[5]
+                    # create an object in the pairtree
+                    pmarc = marc_store.get_object(gale_id, create_if_doesnt_exist=True)
+                    # use pymarc to export as JSON (more compact & faster than XML)
+                    output = StringIO()
+                    writer = pymarc.JSONWriter(output)
```
thatbudakguy, line 43: there's probably little impact to this, but i wonder if things would still work if you moved these outside the forloop. as long as `output` is a buffered IO type, it seems like you could write one record, encode it and flush the buffer, save it to disk, and then repeat without needing to reinitialize the buffer/writer objects

thatbudakguy: and i guess if that logic holds, they could be outside both forloops: you only need one writer for the whole operation, whereas you create one reader per binary marc file

ppa/archive/management/commands/split_marc.py
```diff
@@ -0,0 +1,58 @@
...
+                    pmarc.add_bytestream("marc.json", output.getvalue().encode("utf-8"))
+                    stats["records"] += 1
+
+        self.stdout.write(
+            "Split out %d record%s from %s file%s"
+            % (
+                stats["records"],
+                "" if stats["records"] == 1 else "s",
+                stats["files"],
+                "" if stats["files"] == 1 else "s",
```
thatbudakguy, line 56: this is the exact logic of `pluralize`, no?

ppa/archive/models.py
```diff
@@ -415,6 +416,114 @@ def populate_fields(self, field_data):
...
+            if field_data['publisher'].lower() == '[s.n.]':
+                field_data['publisher'] = ''
+
+            # remove printed by statement before publisher name
+            field_data['publisher'] = re.sub(
+                self.printed_by_re,
+                '',
+                field_data['publisher'],
+                flags=re.IGNORECASE
+            )
```
thatbudakguy, line 501: this seems like an easy candidate for extraction into a separate method, if you wanted to placate codefactor. not sure it matters though

ppa/archive/tests/test_models.py
```diff
@@ -371,6 +371,27 @@ def test_populate_from_bibdata(self):
...
+    def test_metadata_from_marc(self):
+        # majority of logic is tested through test_populate_from_bibdata;
+        # explicitly test only what is different from HathiTrust logic
+
+        # use hathitrust fixture for testing
+        with open(self.bibdata_full) as bibdata:
+            bibdata = hathi.HathiBibliographicRecord(json.load(bibdata))
+
+        digwork = DigitizedWork(source_id='njp.32101013082597')
+        print('year = %s' % bibdata.marcxml.pubyear())
```
thatbudakguy, line 383: leftover print?

thatbudakguy (approved):
using pairtrees to store the data as json seems like a decent solution, and I like that it gives some parity with the hathi data. i might've tried something with json-lines files so that many records could still live in one file, but the ability to go find a single marc record as json on disk seems like it could be handy in the future.

ppa/archive/management/commands/split_marc.py
rlskoeser: Ah, I see how my comment is unclear.

From the pymarc documentation/examples, closing the writer is important to make sure the json is valid. In this case, we need to not close the file handle because that's the output buffer we're writing to.

ppa/archive/management/commands/split_marc.py
rlskoeser: I looked this up for something else recently and the answer I found was that it was more efficient to create a new `StringIO` object than to reset one.

ppa/archive/management/commands/split_marc.py
rlskoeser: yeah, I should just use that

ppa/archive/tests/test_models.py
rlskoeser: oh yeah, thanks

ppa/archive/management/commands/split_marc.py
thatbudakguy: fascinating! I would not have assumed that, but good to know!

rlskoeser:
I thought since we were already using pairtree for the project it seemed like a reasonable and efficient solution.

> ...  i might've tried something with json-lines files so that many records could still live in one file

The ECCO2 marc file is 91MB and ECCO1a and b are 176M and 189M; ECCO2 alone splits out to 48212 files.  Would json lines work at this scale?

It occurred to me after I stopped working on this that binary marc is probably even smaller than json marc, and likely similar or faster to load and parse.  It's not any more difficult to work with, since we're already using pymarc. I'm going to do a little bit of testing to confirm, and may switch it — but I think the way I've written it not too much will have to change.

thatbudakguy:
> The ECCO2 marc file is 91MB and ECCO1a and b are 176M and 189M; ECCO2 alone splits out to 48212 files. Would json lines work at this scale?

I think it's supposed to — one of the ideas is that if you compress the resulting file (e.g. with bzip) there's a ton of redundancy because the JSON structure of each object is repeated, so it plays very well with most compression algorithms. So the size on disk might be considerably smaller than ~150M. Never tried with a file that big though, would be curious to see how it works!

PR #405, opened 2021-06-22
Improve Gale MARC record handling; store and access by ESTC id

rlskoeser:
This PR fixes a problem where MARC records for multivolume works were not being found and also a character encoding problem in some records.

- adds a local lookup file to map Gale Item ID to the ESTC id we need to use to retrieve the MARC records and also volume information (from data provided by Gale, since ESTC and volume are not currently available from the API)
- when splitting MARC records, stores them using the 001 identifier (which is the ESTC id)
- gale import script uses lookup file to get ESTC id for marc record; stores ESTC id and volume information (if any) in the database
- retrieve marc record by record id instead of source id
- specify encoding when loading the marc record from the pairtree (this seems to correct encoding problems)


thatbudakguy (approved):
changes look good; repurposing the `enumcron` for gale records makes sense to me.

Issue #374, opened 2021-06-03
As a user, I want to view a page for content from ECCO in Gale Primary Sources by clicking its thumbnail or page number in a search result so that I can quickly and easily see my search result in its full context.


rlskoeser:
## testing notes
Use keyword search and to get page matches and test clicking on both thumbnail image and page number (they should work the same). Test that it works correctly for both Gale and HathiTrust content, and test on both the main archive search and the search within a single document.

## dev notes
equivalent to #101 for HathiTrust materials; not sure if we know yet how to construct this link, but the Gale folks are eager to help us and will likely respond quickly if we need more information

May require revision to `page_url` template tag, which is currently specific to HathiTrust. If we don't revise, the template tag should be renamed for clarity.

rlskoeser:
Decreasing points from 2 to 1 — this was simple to implement once we get some help from Gale on how to construct the link.

mnaydan:
Tested on main page and search within volume detail page with CW0110416515 and CB0130649164. Thumbnail and page number links work!

Majority of HT items I tested work as expected on main and detail pages. Of course I stumbled upon a needle in the haystack: for wu.89003553799, the displayed snippet and page number are correct, but displayed thumbnail and links are not. Search "bruce and the spider" and it displays/links to reader page 19 (actual page 17). It should display/go to reader page 17 (actual page 15). 

But I can't find the same problem in any other volume I've tested. @rlskoeser perhaps you could quickly investigate before I close, but it could just be a fluke of the volume, I guess? I'll keep an eye on whether it could be related to the page range changes for excerpts, too...

mnaydan:
Also just a note that frontmatter and backmatter is still not searchable. Bill had said that we could expect the API update possibly by July 25 or "near" July 2, so I'm not sure whether they are behind or we need to update something on our end.

rlskoeser:
@mnaydan regarding front matter & back matter: we will have to re-import to get the front and back matter OCR. We should check the API first to see if the ocr text is available yet, and then we can do a test import of one of the volumes to confirm. Are you able to check the API or should I ?

rlskoeser:
> Majority of HT items I tested work as expected on main and detail pages. Of course I stumbled upon a needle in the haystack: for wu.89003553799, the displayed snippet and page number are correct, but displayed thumbnail and links are not. Search "bruce and the spider" and it displays/links to reader page 19 (actual page 17). It should display/go to reader page 17 (actual page 15).

[...]

PR #406, opened 2021-06-23
Add & test logic to link to specific pages on Gale

rlskoeser:
ref #374

- rename existing `page_url` template tag to `hathi_page_url`
- add new, equivalent `gale_page_url` template tag to generate page-level links for gale content

### DIFF
--- a/ppa/archive/models.py
+++ b/ppa/archive/models.py
@@ -934,6 +934,10 @@ class Page(Indexable):
         making an extra API call if data is already available."""
         if gale_record is None:
             gale_record = GaleAPI().get_item(digwork.source_id)
+        # NOTE when adding support for excerpts from Gale, we should
+        # ensure that order is set based on the sequence of images within the volume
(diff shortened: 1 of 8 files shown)


thatbudakguy (approved):
everything seems fine here; only thought is that I'd keep `page_url` as the templatetag and then just have it detect and delegate to the relevant function depending on if the item is Gale or Hathi. that way you save on some template logic, but can keep the current tests.

rlskoeser:
> everything seems fine here; only thought is that I'd keep `page_url` as the templatetag and then just have it detect and delegate to the relevant function depending on if the item is Gale or Hathi. that way you save on some template logic, but can keep the current tests.

@thatbudakguy that was my initial thought as well, but they take different arguments — the HathiTrust page url function takes an item id + order, but the Gale function needs item _url_. Maybe I could rework the HathiTrust logic and indexing to use an item view url instead of id to make them more consistent that way, but that seems like a bigger change and could potentially introduce new problems.

PR #407, opened 2021-06-24
Add fields for excerpt support and configure in admin

rlskoeser:
Includes validation and testing for page range formatting

Now includes functionality to limit pages that are indexed based on the page range, in addition to adding the fields and admin interface to manage the fields needed to support excerpts.

If a work is saved and the page range has changed, it will recalculate page count based on the range and update page indexing. I don't think it will work exactly right for _removing_ a page range and going back to a full item, but I think it's ok not to support that.


rlskoeser:
I figured out a way to make this actionable — setting a page range in admin will now trigger page count recalculation (based on the range) and is also used to filter the pages that are indexed. Hopefully this gives us a light-weight way to start testing excerpt functionality independent of the import features

ppa/archive/models.py
```diff
@@ -11,10 +11,13 @@
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
+from django.core.validators import RegexValidator
 from django.db import models
 from django.urls import reverse
 from eulxml.xmlmap import load_xmlobject_from_file
 from flags import Flags
+from intspan import ParseError as IntSpanParseError
+from intspan import intspan
```
thatbudakguy, line 20: wow, this is great! i feel like we encounter applications for this all the time; "page range" seems to come up often

thatbudakguy (approved):
everything looks good here — do we indicate anywhere in the admin UI that removing page ranges isn't supported, though? that seems important to note (although it probably will rarely if ever occur).

rlskoeser:
> everything looks good here — do we indicate anywhere in the admin UI that removing page ranges isn't supported, though? that seems important to note (although it probably will rarely if ever occur).

good point, I should probably add a warning to the help text or somewhere like that

ppa/archive/models.py
rlskoeser: yeah, I was really pleased when I found it, makes our code much easier

Issue #396, opened 2021-06-21
As an admin, I want the option of importing two different sections from the same HathiTrust work so I can include multiple articles or chapters from a single journal issue or book.


rlskoeser:
## testing notes
You should be able to test this functionality by changing existing records and adding a new one. 
- Choose an existing non-excerpt digitized work to test with. Make a note of the public url for this work because you are going to need it for testing.
-  Edit the digitized work you chose to turn it into an excerpt (for the purposes of this test, the only field that really matters is the **page range (digital edition)**). For convenience, use "Save and continue editing."
- [x] Use the "view on site" button to view the public version of the item you modified. Check that the url is based on the source id *plus* the first page number of the excerpt.
- [x] Test the public url of this work that you saved before converting it to an excerpt: it should redirect to the new url with the page number.
- Manually add another excerpt from the same source. (For the purposes of this test, the fields that matter are **source id** and **page range (digital edition)**). Save the record.
- [x] Test the public url of this work that you saved before converting it to an excerpt: it should now return a 404 not found, since there is more than one excerpt from the original work.

## dev notes
- [x] Revise database unique requirements for `DigitizedWork`  — previously source id was required to be unique; will need to be  the combination of source id and page sequence that is required to be unique together.
- [x] Revise work identifier in index data so that different articles from the same source can be distinguished
- [x] revise url for public document detail page so it uses source id + start page for excerpts
- [x] implement see other redirect from old url to excerpt url

mnaydan:
@rlskoeser I don't know if this messes you up but it is not always just 2. I did a quick check and I _think_ the most we have would be 3.

rlskoeser:
@mnaydan question for you while I'm thinking about it (in relation to our recent conversation on #393, about converting existing records from full to excerpt or reimporting) — if we need to allow multiple excerpts from the same source, then I will need to change the URLs on the PPA site to make sure they are unique. (Right now they are based on source id.)  

I was thinking of using source id combined with starting page number for urls for excerpts. Can you think of any case where that would not be unique? Also, can you think of any reason the starting page number would change after you've added something (i.e., would this be a stable identifier)? If you have any ideas for something different, let me know.

mnaydan:
@rlskoeser oh, good point. That makes sense to me. In one case, the end page of one excerpt is the same as the start page of the next excerpt (e.g., 459-466 and 466-479) - but the starting pages are different, so they'd still be unique. After looking over the cases I have, I think it should work.

[...]

PR #409, opened 2021-07-09
Handle multiple excerpts from the same source; update indexing and detail url

rlskoeser:
per #396 : 
- Revise database unique requirements for DigitizedWork (previously source id was required to be unique; now combination of source id and page sequence should be unique together).
- Revise work identifier in index data so that different articles from the same source can be distinguished and pages can be grouped correctly
- revise url for public document detail page to uses source id + start page for excerpts
- implement see other redirect from old full-work url to excerpt url if there is a single excerpt


ppa/archive/models.py
```diff
@@ -389,13 +389,24 @@ class DigitizedWork(TrackChangesModel, ModelIndexable):
...
         """
-        return reverse("archive:detail", kwargs={"source_id": self.source_id})
+        url_opts = {"source_id": self.source_id}
+        # start page must be specified if set but must not be included if empty
+        if self.pages_digital:
+            url_opts["start_page"] = self.first_page()
+        return reverse("archive:detail", kwargs=url_opts)
 
     def __str__(self):
         """Default string display. Uses :attr:`source_id`"""
```
thatbudakguy, line 412: should this incorporate the page range if set? not a big deal, just seemed like it might make sense

ppa/archive/models.py
```diff
@@ -707,9 +725,12 @@ def index_data(self):
         if self.status == self.SUPPRESSED:
             return {"id": self.source_id}
 
+        index_id = self.index_id()
         return {
-            "id": self.source_id,
+            "id": index_id,
             "source_id": self.source_id,
+            "first_page_i": list(self.page_span)[0] if self.pages_digital else None,
```
thatbudakguy, line 732: can this just call `first_page()`? seems like the same logic

ppa/archive/solr.py
```diff
@@ -126,6 +126,7 @@ class ArchiveSearchQuerySet(SolrQuerySet):
         "collections",
         "source_t",
         "image_id_s",
+        "first_page_i",
```
thatbudakguy, line 129: does `group_id` not get added as well?

ppa/archive/views.py
```diff
@@ -249,20 +256,59 @@ class DigitizedWorkDetailView(AjaxTemplateMixin, SolrLastModifiedMixin, DetailVi
...
-        response = super().get(*args, **kwargs)
+        try:
+            response = super().get(*args, **kwargs)
+        except Http404:
+            # if see other url is set (i.e., tried to retrieve a non-existent
+            # full work, but there is one excerpt with that source id)
+            if self.see_other_url:
+                return HttpResponseSeeOther(self.see_other_url)
+            # otherwise, let the 404 propagate
+            raise
```
thatbudakguy, line 306: this is a strange situation...I don't know if 303 makes sense here, but I'm also not sure anything else does. presuming we're implementing this mostly so that any links generated prior to excerpt support which appeared to point to full works will now do the right thing and point to the excerpt. [based on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303), it looks like 303 is usually used in response to POST/PUT to indicate that the server is responding with a "you successfully uploaded X" message rather than the object you actually uploaded, which isn't our use case at all. [the actual spec](https://datatracker.ietf.org/doc/html/rfc7231#section-6.4.4) is a lot more vague, but it also only really mentions that use case.

would it be a problem to just permanently redirect with a 301? it seems like if we are including excerpts of an item in the PPA, there's no scenario in which we would also include the entire item, so it's fine to permanently redirect to the single excerpt (until such time as another excerpt is added, at which point the request is too ambiguous to resolve, so we just 404).

thatbudakguy (approved):
few implementation Qs but code looks good. I'm not opposed to usage of HTTP 303, it just seems like the semantics of it are not what I'd expect judging from the name.

ppa/archive/views.py
thatbudakguy: very tangential: if we do use 303 and are happy with the semantics, maybe we should also use [HTTP 300](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/300) for the case where there are multiple excerpts!

ppa/archive/models.py
rlskoeser: yes, good idea

ppa/archive/models.py
rlskoeser: good catch. (I'm gonna say the method didn't exist yet when I first wrote this .... but who knows)

ppa/archive/solr.py
rlskoeser: I don't think we need it in the returned result, only server side for grouping works and pages

ppa/archive/views.py
rlskoeser: Hmm, yeah. I was thinking it shouldn't be a 301 because it isn't the same resource. I _think_ any client that doesn't care about the specifics will just see it as a redirect; but maybe the 301 is closer to what we want, if what we care about most is bookmarks and indexing.  I don't know what search engines do with 303s — [this article](https://www.seoclarity.net/pillar/url-redirects) indicates it is treated like a temporary redirect; I see other posts and discussions recommending not to use it for SEO reasons.

I considered 300, actually! From what I read, there's no standard way to generate the choice list so we'd have to generate an html page — and I wasn't sure it was worth that much effort. (Although now that I think more about it, maybe it wouldn't be that hard; maybe we could adapt our error page template and just pass in a list of choices... but still think this is a nice-to-have and we're tight on time)

PR #410, opened 2021-07-12
Index work item type in solr and include in results for use in displaying article/excerpt icon

rlskoeser:
also switched to aliased solr queryset so we can provide more readable names for solr dynamic fields

### DIFF
--- a/ppa/archive/models.py
+++ b/ppa/archive/models.py
@@ -753,6 +753,9 @@ class DigitizedWork(TrackChangesModel, ModelIndexable):
             # hard-coded to distinguish from & sort with pages
             "item_type": "work",
             "order": "0",
+            "work_type_s": self.get_item_type_display()
+            .lower()
+            .replace(" ", "-"),  # full, excerpt, or article
         }
 
     def remove_from_index(self):
--- a/ppa/archive/solr.py
+++ b/ppa/archive/solr.py
@@ -1,7 +1,7 @@
 import logging
 
 from parasolr import schema
-from parasolr.django import SolrQuerySet
+from parasolr.django import AliasedSolrQuerySet, SolrQuerySet
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


ppa/archive/solr.py
```diff
@@ -128,11 +128,26 @@ class ArchiveSearchQuerySet(SolrQuerySet):
...
+        "work_type_s",
     ]
+    # aliases for any fields we want to rename for search and display
+    # (must also be included in return_fields list)
+    aliases = {
+        "source_t": "source",
+        "image_id_s": "image_id",
+        "first_page_i": "first_page",
+        "work_type_s": "work_type",
+    }
```
thatbudakguy, line 140: i think some of these get referenced in other open PRs so the merge order might matter (unless you can still access a field via its original un-aliased name)

ppa/archive/models.py
```diff
@@ -753,6 +753,9 @@ def index_data(self):
             # hard-coded to distinguish from & sort with pages
             "item_type": "work",
             "order": "0",
+            "work_type_s": self.get_item_type_display()
+            .lower()
+            .replace(" ", "-"),  # full, excerpt, or article
```
thatbudakguy, line 758: why the `replace()`? comment suggests there won't ever be values that have whitespace

thatbudakguy (approved):

ppa/archive/models.py
rlskoeser: the actual names are "Full Work", "Excerpt", and "Article"

rlskoeser: I'm open to you revising this when you implement the icons if you have a different solution you prefer! I wondered if it would matter for the accessibility label you were thinking of adding — but I don't think we need to label/announce full works.

ppa/archive/solr.py
rlskoeser: thanks, I'll keep an eye on that when I merge — I think this one should go first but will be good to confirm!

ppa/archive/models.py
thatbudakguy: oh, makes sense! I figured it was to make it work as a class name. can you change the comment to say "full-work" so it shows the real value that will be used?

PR #413, opened 2021-07-13
new manage command to convert HathiTrust digitized works into excerpts

rlskoeser:
manage command expects a csv with details for the excerpts and does the following:

- if a full work is found for the source id, the existing record is converted to an excerpt
- if not (i.e., multiple excerpts for the source and it's already been excerpted), create a new record; assumes that HathiTrust data is already available 
- whether new or existing record, populate all fields from the spreadsheet
- set collection membership
- create a log entry documenting creation or excerption
- summarize what was done

Currently the only error handling is for required fields in the csv and parse errors in the page range (which might be enough).

### DIFF
--- a/ppa/archive/management/commands/hathi_excerpt.py
+++ b/ppa/archive/management/commands/hathi_excerpt.py
@@ -27,20 +27,26 @@ class Command(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument("csv", help="CSV file with excerpt information")
 
+    def setup(self):
+        # common setup steps for running the script or testing
+        self.stats = Counter()
+        self.script_user = User.objects.get(username=settings.SCRIPT_USERNAME)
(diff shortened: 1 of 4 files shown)


ppa/archive/management/commands/hathi_excerpt.py
```diff
@@ -0,0 +1,165 @@
...
+        excerpt_info = self.load_csv(kwargs["csv"])
+        # load collections from the database
+        self.load_collections()
+        self.excerpt(excerpt_info)
+
+        self.stdout.write(
+            "\nExcerpted {excerpted:,d} existing records; created {created:,d} new excerpts. {error:,d} errors.".format_map(
+                self.stats
+            )
+        )
```
thatbudakguy, line 49: TIL about `format_map()`, cool!

thatbudakguy (approved):
everything looks great!

Issue #397, opened 2021-06-21
As a user, when I'm looking at search results I want to see an indicator when something is an excerpt or an article, so that I can tell what kind of content I'm looking at.


rlskoeser:
## testing notes
**design links**
[excerpt + article icons](https://zpl.io/VQJdWpA)
- [x] excerpt should have excerpt icon
- [x] article should have article icon
- [x] book/journal title should display in place of full work subtitle

## dev notes
- [x] include item type in work index data
- [x] display excerpt and article icons for items as appropriate
- [x] include book or journal title in search results  _only if_ we can easily integrate it into current format/design

mnaydan:
Placement looks good across browsers and devices. Book/journal title display in subtitle area looks GREAT. Two things:

1) The lines on the book excerpt icon appear distorted, especially on Chrome and FF (screenshot 1) and also when the icon is mobile size (screenshot 2). Perhaps the icon needs to be simplified? 

2) On all browsers if window is sized between full and tablet, icons get cut off (screenshot 3). I think it should either snap to tablet view earlier, or include some left margin padding?

Both of these issues are happening on detail view, too. 

Screenshot 1:
<img width="133" alt="Screen Shot 2021-07-16 at 12 57 13 PM" src="https://user-images.githubusercontent.com/22945600/126341765-b6a43fe4-4d82-4ca1-8be2-2afddf3fad34.png">

Screenshot 2:
<img width="66" alt="Screen Shot 2021-07-20 at 10 11 41 AM" src="https://user-images.githubusercontent.com/22945600/126341773-1b75245a-6ff6-40b5-bee1-304c2143a1f8.png">

Screenshot 3:
<img width="965" alt="Screen Shot 2021-07-20 at 10 12 21 AM" src="https://user-images.githubusercontent.com/22945600/126342045-16286fb9-88f2-4a31-bbb5-5609f8e8a177.png">

mnaydan:
Sorry for the multiple comments. Another couple general things I noticed while testing the other issue:

Sometimes the right line on the article icon disappears. This happens inconsistently to me when browsing. Perhaps due to window size?: 
<img width="945" alt="Screen Shot 2021-07-20 at 10 42 14 AM" src="https://user-images.githubusercontent.com/22945600/126354051-cc302a37-8bff-412a-b85b-44c4cbe50f4f.png">

[...]

PR #414, opened 2021-07-14
397 excerpt search ui

thatbudakguy:
- Excerpt/article icon on search results (#397)
- Add excerpt/article book/journal display on search results

### DIFF
--- a/ppa/archive/templates/archive/snippets/search_result.html
+++ b/ppa/archive/templates/archive/snippets/search_result.html
@@ -14,6 +14,7 @@
                     {% if page_obj %}
                         <p class="result-number">{{ page_obj.start_index|add:forloop.counter0 }}</p>
                     {% endif %}
+                    <div class="work-type" aria-label="{{ item.work_type }}"></div>
                 </div>
                 <p>
                     <span class="ui header">{{ item.title }}</span>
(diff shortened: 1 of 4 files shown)


ppa/archive/templates/archive/snippets/search_result.html
```diff
@@ -14,10 +14,17 @@
                     {% if page_obj %}
                         <p class="result-number">{{ page_obj.start_index|add:forloop.counter0 }}</p>
                     {% endif %}
+                    <div class="work-type" aria-label="{{ item.work_type }}"></div>
```
rlskoeser, line 17: does this make sense for full works? or could we just omit it for for full work?

rlskoeser: I tried testing the search results page with voice over because I wanted to hear what it sounded like, and I couldn't tab through the search form into the results — it goes from the sort by option back to the browser back button.  Am I doing something wrong?

rlskoeser (approved):
Looks great! One question about aria label and an accessibility concern that isn't related to your changes (but we should address separately and quickly if at all possible!)

rlskoeser:
I see that you've fixed the item type aria label so it is only included for non full works in your other branch, so I'm going to go ahead and merge these.

Issue #398, opened 2021-06-21
As a user, when I'm looking at the details for an item I want to see an indicator if it's an excerpt or article so I understand the content better.


rlskoeser:
## testing notes
**design links**
[excerpt + article icons](https://zpl.io/VQJdWpA)
- [x] article icon should display for articles
- [x] excerpt icon should display for excerpts
- [x] book or journal title should display when set, with appropriate title based on type
- [x] original page range should display when set
- [x] link to item in HathiTrust should go to the first page in range 

## dev notes
- [x] include excerpt/article icons on document detail page based on item type
- [x] include additional metadata on document detail page based on item type: book or journal title, original page range
- [x] link to view item in HathiTrust or Gale should link to the first page in sequence for the excerpt

mnaydan:
Everything above looks good, love how journal/book title on detail view changes automatically depending on content type. Same icon problems as I noted on #397 but we can track them there.

FYI: The HT link to first page works on frontend, but on backend it doesn't. But it works "as a user," so I can track that as an admin issue elsewhere...

PR #415, opened 2021-07-14
398 excerpt detail ui

thatbudakguy:
- Excerpt/article icon on search results (#397)
- Add excerpt/article book/journal display on search results
- Add excerpt/article icons to detail view (#398)
- Add page range display and update view links for excerpts/articles

### DIFF
--- a/ppa/archive/templates/archive/digitizedwork_detail.html
+++ b/ppa/archive/templates/archive/digitizedwork_detail.html
@@ -106,6 +106,7 @@
                         {% if object.source == object.HATHI %}
                         <a href="{% hathi_page_url object.source_id object.first_page %}" target="_blank" rel="noopener noreferrer">{{ object.source_id }}</a>
                         {% elif object.source == object.GALE %}
+                        GALE url
                         <a href="{% gale_page_url object.source_url object.first_page %}" target="_blank" rel="noopener noreferrer">{{ object.source_id }}</a>
                         {% endif %}
                     {% else %} {# when there is no page range, use source url #}
--- a/ppa/archive/templates/archive/digitizedwork_detail.html
+++ b/ppa/archive/templates/archive/digitizedwork_detail.html
@@ -45,6 +45,12 @@
                 <td>{{ object.subtitle }}</td>
             </tr>
             {% endif %}
+            {% if object.book_journal %}
+            <tr class="book-journal">
+                <th scope="row">{% if object.item_type == "A" %}Journal{% elif object.item_type == "E" %}Book{% endif %} Title</th>
+                <td>{{ object.book_journal }}</td>
+            </tr>
+            {% endif %}
             {% if object.pub_date %}
             <tr>
                 <th scope="row">Publication Date</th>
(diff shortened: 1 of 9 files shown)


ppa/archive/templates/archive/digitizedwork_detail.html
```diff
@@ -32,6 +32,9 @@
 
 {% block content %}
 <section class="meta ui basic vertical segment container">
+    {% if object.item_type != 'F' %}
```
rlskoeser, line 35: this check needs to use the constants on the class, not hard-coded strings

ppa/archive/templates/archive/digitizedwork_detail.html
```diff
@@ -42,6 +45,12 @@ <h1 class="header">{{ object.title }}</h1>
                 <td>{{ object.subtitle }}</td>
             </tr>
             {% endif %}
+            {% if object.book_journal %}
+            <tr class="book-journal">
+                <th scope="row">{% if object.item_type == "A" %}Journal{% elif object.item_type == "E" %}Book{% endif %} Title</th>
```
rlskoeser, line 50: same comment about the constants

I wonder if this should be a method on the class, but IDK if we'll need it anywhere else

ppa/archive/templates/archive/digitizedwork_detail.html
```diff
@@ -87,7 +102,13 @@ <h1 class="header">{{ object.title }}</h1>
             <tr class="source-link">
                 <th scope="row">{{ object.get_source_link_label }}</th>
                 <td>
+                    {% if object.source == "G" %}
```
rlskoeser, line 105: same problem with the check here — need to either use the constants on the class or methods on the class (if we have them, but I don't think we do and it doesn't make sense to add them)

Also, we should only use the page url if there are pages defined

ppa/archive/templates/archive/snippets/search_result.html
```diff
@@ -14,10 +14,19 @@
                     {% if page_obj %}
                         <p class="result-number">{{ page_obj.start_index|add:forloop.counter0 }}</p>
                     {% endif %}
+                    {% if item.work_type != "full-work" %}
+                    <div class="work-type icon" aria-label="{{ item.work_type }}"></div>
+                    {% endif %}
                 </div>
```
rlskoeser, line 20: 👍  thanks for correcting

ppa/archive/templates/archive/snippets/search_result.html
```diff
@@ -14,10 +14,19 @@
...
                 <p>
+                    {% spaceless %}
                     <span class="ui header">{{ item.title }}</span>
+                    {% if item.work_type != "full-work" and item.book_journal %}
```
rlskoeser, line 24: I think the extra check for type _shouldn't_ be necessary, but right now we don't have much validation on the fields for the different types so I see the value in keeping it

rlskoeser (changes requested):
some minor cleanup needed

rlskoeser (approved):

PR #416, opened 2021-07-14
small fixes for hathi excerpt script

rlskoeser:
minor update with some small fixes based on what @mnaydan found in testing so far

@thatbudakguy interested in your thoughts on the excerpt url problem: the syntax I'm using now _doesn't_ error if it fails, so it wasn't obvious this was a problem.

### DIFF
--- a/ppa/archive/templates/archive/snippets/search_result.html
+++ b/ppa/archive/templates/archive/snippets/search_result.html
@@ -2,8 +2,8 @@
 
 <li class="item {{ item.work_type }}">
     <abbr class="unapi-id" title="{{ item.source_id }}" aria-hidden="true"></abbr>
-    {% if item.first_page_i %}
-        {% url 'archive:detail' source_id=item.source_id start_page=item.first_page_i as item_url %}
+    {% if item.first_page %}
+        {% url 'archive:detail' source_id=item.source_id start_page=item.first_page as item_url %}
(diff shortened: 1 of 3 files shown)


ppa/archive/templates/archive/snippets/search_result.html
```diff
@@ -2,8 +2,8 @@
 
 <li class="item {{ item.work_type }}">
     <abbr class="unapi-id" title="{{ item.source_id }}" aria-hidden="true"></abbr>
-    {% if item.first_page_i %}
-        {% url 'archive:detail' source_id=item.source_id start_page=item.first_page_i as item_url %}
+    {% if item.first_page %}
+        {% url 'archive:detail' source_id=item.source_id start_page=item.first_page as item_url %}
```
rlskoeser, line 6: This stopped working correctly when I changed the solr aliases, but it didn't generate any errors — I was aware of the fact that that `url` doesn't error when you use `as` where it would error if you don't assign it to a variable, but I thought it was better not to have to repeat the url logic and use a variable.

Options:
- drop the variable and just explicitly generate the url where we need it, without setting a variable (we don't even use it twice! — a little less readable, but will cause an error if there's a problem)
- add a unit test to make sure we're generating excerpt urls correctly in the template (probably should do this no matter what)
- add athat sets the url

PR #421, opened 2021-07-27
Revise hathi excerpt indexing and TrackChangesModel logic

rlskoeser:
- `TrackChangesModel.has_changes()` now returns false for unsaved models
- re-ordered indexing changes in `DigitizedWork.save()` so that clearing index due to a changed group id happens _before_ indexing pages due to a changed page range
- adjusted `hathi_excerpt` script to count and index pages for newly created excerpts (previously happening on save due to the old `has_changes()` logic, but not logical that it should)


thatbudakguy (approved):
looks good!

Issue #474, opened 2022-10-24
add test matrix to CI; test against: postgres + solr 8; mysql + solr 6


rlskoeser:

rlskoeser:
confident this is working as desired, now that we have the site running successfully in the PUL test environment

PR #479, opened 2022-10-24
Fixing github actions for migration

quadrismegistus:
This is a fix to #474.


quadrismegistus:
Notes as I work. Here are the steps I've followed, from scratch:

* Clone repository, follow README

* Set up mysql

```
mysql -u root -e "DROP USER ppa;"
mysql -u root -e "DROP DATABASE ppa;"
mysql -u root -e "CREATE USER 'ppa' IDENTIFIED BY 'ppa';"
mysql -u root -e "GRANT ALL PRIVILEGES ON *.* TO 'ppa' WITH GRANT OPTION;"
mysql --user=ppa --password=ppa -e "CREATE DATABASE ppa;"
```

* Set up psql:

```
psql -d postgres -c "DROP DATABASE ppa;"
psql -d postgres -c "DROP ROLE ppa;"
psql -d postgres -c "CREATE ROLE ppa WITH CREATEDB LOGIN PASSWORD 'ppa';"
psql -d postgres -U ppa -c "CREATE DATABASE ppa;"
```

* Edit local_settings.py, initially setting mysql to default:

```
DATABASES = {
    'default': {
        "ENGINE": "django.db.backends.mysql",
        'NAME': 'ppa',
        'USER': 'ppa',
        'PASSWORD': 'ppa',
    },
    "postgres": {
        "ENGINE": "django.db.backends.postgresql",
        "NAME": "ppa",
        "USER": "ppa",
        "PASSWORD": "ppa",
        "HOST": "",  # empty string for localhost
        "PORT": "",  # empty string for default
    }
}
```

* Migrate into django:

```
python manage.py migrate
```

* Load data from mysql dump

```
mysql --user=ppa --password=ppa --database=ppa < data/2022-10-04-ppaprod.sql   # Or similar path
```

* Dump this data in jsonl format:

```
python manage.py dumpdata --all --format jsonl --output data/ppa_dumpdata.jsonl
```

* Switch postgres to default database in local_settings.py

```
DATABASES = {
    'mysql': {
        "ENGINE": "django.db.backends.mysql",
        'NAME': 'ppa',
        'USER': 'ppa',
        'PASSWORD': 'ppa',
    },
    "default": {
        "ENGINE": "django.db.backends.postgresql",
        "NAME": "ppa",
        "USER": "ppa",
        "PASSWORD": "ppa",
        "HOST": "",  # empty string for localhost
        "PORT": "",  # empty string for default
    }
}
```

* Import data into django (now postgres):

```
python manage.py loaddata data/ppa_dumpdata.jsonl 
```

quadrismegistus:
I've edited the unit-tests.yml file to run a whole separate set of tests (duplicated) for the separate config of postgres + solr8. This may be too much redundancy? But right now this is what's killing the tests, this error recurs again and again in test output:

```
self = <django.db.backends.postgresql.creation.DatabaseCreation object at 0x7f31b1f38f98>

    def sql_table_creation_suffix(self):
        test_settings = self.connection.settings_dict['TEST']
        assert test_settings['COLLATION'] is None, (
>           "PostgreSQL does not support collation setting at database creation time."
        )
E       AssertionError: PostgreSQL does not support collation setting at database creation time.

/opt/hostedtoolcache/Python/3.6.9/x64/lib/python3.6/site-packages/django/db/backends/postgresql/creation.py:25: AssertionError
```

Not immediately sure how to fix this error but this is where I should probably leave it for today. Will come back to it tmrw

quadrismegistus:
Okay I believe all these tests are working now. The solution was to use the matrix options to test both mysql/solr6 + psql/solr8. The `unit-tests.yml` in `mep-django` was used as model. However, python3.5 as used in the former raised deprecation errors and so python 3.6 was used for both tests.

.github/workflows/unit-tests.yml
```diff
@@ -2,12 +2,15 @@ name: unit tests
 
 on:
   push: # run on every push or PR to any branch
-  pull_request:
+  # pull_request: # for now to stop duplication?
```
rlskoeser, line 5: We will want it enabled on PRs going forward; maybe re-enable before merging?

.github/workflows/unit-tests.yml
```diff
@@ -31,52 +34,128 @@ jobs:
...
+
+      - name: Copy solr configsets to solr home directory (Solr 8 only)
+        run: "docker exec -d ${{ job.services.solr.id }} cp -r /opt/solr/server/solr/configsets /var/solr/data"
+        if: ${{ matrix.solr == 8.6 }}
+
+      # NOTE: for duration of migration to PUL, use python version from build matrix
+      # Python version to use is stored in the .python-version file, which is the
+      # convention for pyenv: https://github.com/pyenv/pyenv
+      # - name: Get Python version
+      #   run: echo "PYTHON_VERSION=$(cat .python-version)" >> $GITHUB_ENV
```
rlskoeser, line 95: we aren't changing the python version, are we? same for both? I would use it here and not in the matrix, but may not matter so much

.github/workflows/unit-tests.yml
```diff
@@ -31,52 +34,128 @@ jobs:
...
+          pip install -r dev-requirements.txt
+
+      # This unnec?
+      # - name: Download webpack stats
+      #   uses: actions/download-artifact@v2
+      #   with:
+      #     name: webpack-stats.json
+
+      # - name: Debug output database configuration
+      #   run: python -c "import pprint; import ci.testsettings; pprint.pprint(ci.testsettings.DATABASES)"
```
rlskoeser, line 125: I can't remember what this was for! I think maybe a shared dependency for  the python tests and js tests... ?

Disabling it didn't cause any problems or warnings?

.github/workflows/unit-tests.yml
```diff
@@ -31,52 +34,128 @@ jobs:
...
+        uses: rtCamp/action-slack-notify@master
         if: ${{ always() && (github.event_name == 'schedule' || github.event_name == 'push') }}
         continue-on-error: true
         env:
           SLACK_COLOR: ${{ env.SLACK_COLOR }}
           SLACK_WEBHOOK: ${{ secrets.ACTIONS_SLACK_WEBHOOK }}
-          SLACK_TITLE: "Workflow `${{ github.workflow }}`: ${{ job.status }}"
-          SLACK_MESSAGE: "Run <https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}|#${{ github.run_number }}> on <https://github.com/${{ github.repository }}/|${{ github.repository }}@${{ github.ref }}>"
+          SLACK_TITLE: "Walk #${{ github.run_number }} for workflow `${{ github.workflow }} (python${{ matrix.python }}/${{ matrix.db }}/solr${{ matrix.solr }})`: ${{ job.status }}"
+          SLACK_MESSAGE: "Walk <https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}|#${{ github.run_number }}> on <https://github.com/${{ github.repository }}/|${{ github.repository }}@${{ github.ref }}>"
```
rlskoeser, line 159: 😂 🚶

ci/testsettings-psql.py
```diff
@@ -0,0 +1,50 @@
+# This file is exec'd from settings.py, so it has access to and can
```
rlskoeser, line 1: is this one needed? I don't see it referenced in the build

DEVELOPERNOTES.rst
```diff
@@ -26,7 +26,7 @@ Create a new core with the `ppa` configset::
 
 When the configset has changed, copy in the updated solr config files::
 
-    docrker cp solr_conf/* solr6:/opt/solr/server/solr/configsets/ppa/
+    docker cp solr_conf/* solr6:/opt/solr/server/solr/configsets/ppa/
```
rlskoeser, line 29: 🤦‍♀️ thank you for cleaning up after me

rlskoeser (commented):
This looks great! I think only a few small changes before we can merge.

It looks like you didn't have to change any of the solr queries to get the tests passing... I'm wondering if that means the tests don't cover the cases that would be impacted or if we already updated it for solr8 at some point in the past when we knew what the change would be.

I guess we can keep an eye out for that when we test functionality on the new test site.

rlskoeser:
@quadrismegistus forgot to say: if there are any steps we need to remember to do after the migration to cleanup the build, you could start an issue now and put some todos.  (If it's just removing the matrix maybe not worth it, but if there is anything else could be worth making some notes)

.github/workflows/unit-tests.yml
quadrismegistus: I believe we initially were -- I should have noted this -- using Python 3.5 for the mysql + solr6 configuration, but I was getting an error that stackoverflow suggested was an artifact of 3.5 vs 3.6, so I switched to 3.6 for the time being. Let me know what you want to do – I can try to get 3.5 to work tho I also think 3.6 is reasonable to expect at this point?

ci/testsettings-psql.py
quadrismegistus: woops no this is left over from an earlier solution

.github/workflows/unit-tests.yml
rlskoeser: oh! my apologies, that's another leftover from the mep code I pointed you to — in that case we were still on 3.5 but migrating to 3.6, in ppa we're already on 3.6. Let's restore this bit and remove python version from the matrix

.github/workflows/unit-tests.yml
rlskoeser: Whew, makes me feel better not remembering it. Sounds like it's fine to assume it doesn't apply here

rlskoeser (approved):
looks great, just remember to re-enable running on PRs before you merge

Issue #475, opened 2022-10-24
convert solr schema configuration from schema api fields to configset


rlskoeser:

rlskoeser:
We were able to setup the test solr core in PUL test infrastructure with the new ansible playbook, indexed content into Solr, and it seems like searching the public interface is working as expected — so I think this can be considered done also.

PR #480, opened 2022-11-02
Remove outdated solr schema configuration and instructions

rlskoeser:
ref #475

It looks like the solr schema was already previously updated to use the managed schema file and not the schema api, and there was just some old code and instructions left hanging around that we don't need anymore. I think if all the tests pass it's probably safe to jettison this stuff (I spot-checked some of the fields, field types, and copy fields and found them in the managed schema).


quadrismegistus (approved):
I guess in some ways the best commits are deletions. Looks good!

Issue #482, opened 2022-11-08
Progress bar breaking on solr indexing when count of pages is off


quadrismegistus:

quadrismegistus:
Going to just fix this by setting max_error to be False on pbar.ProgressBar:

```
class ProgressBar(StdRedirectMixin, ResizableMixin, ProgressBarBase):
    '''The ProgressBar class which updates and prints the bar.

    Args:
        min_value (int): The minimum/start value for the progress bar
        max_value (int): The maximum/end value for the progress bar.
                            Defaults to `_DEFAULT_MAXVAL`
...
        max_error (bool): When True the progressbar will raise an error if it
            goes beyond it's set max_value. Otherwise the max_value is simply
            raised when needed
```

rlskoeser:
Tested indexing pages yesterday multiple times on the new test vm server, and never got the page count error we saw when we first ran it after the pgloader migration.

I also experimented a little bit with the number of processes. Our test VM has 2 core, and the script uses the same number of processes as cores by default — when I ran with 4 processes instead of 2 it was a bit faster. I _suspect_ this is because indexing Gale content requires waiting on their API, which can be slow for larger texts.

Here is the `time` output for a couple of different runs:

Indexing with default 2 processes:
```
Items in Solr by item type: 1897884 pages, 6456 works

real    95m13.133s
user    45m49.996s
sys     0m28.560s
```

Indexing with 4 processes:
```
Items in Solr by item type: 1897884 pages, 6456 works

real    50m1.842
user    53m15.275s
sys     0m42.020s
```
I tested with 8 processes just to see if it made any difference, and it was about the same as the time for 4 processes.

rlskoeser:
closing since the progress bar is no longer broken

PR #483, opened 2022-11-08
Fix to progress bar when max_value set too low

quadrismegistus:
This is a fix for https://github.com/Princeton-CDH/ppa-django/issues/482

### DIFF
--- a/ppa/archive/management/commands/generate_corpus.py
+++ b/ppa/archive/management/commands/generate_corpus.py
@@ -169,7 +169,11 @@ class SolrCorpus:
         self.doc_ids = self.page_counts.keys()
         self.doc_count = len(self.doc_ids)
         if pbar:
-            self.pbar = ProgressBar(redirect_stderr=True, max_value=self.doc_count)
+            self.pbar = ProgressBar(
+                redirect_stderr=True, 
+                max_value=self.doc_count, 
+                max_error=False
+            )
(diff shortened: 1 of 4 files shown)


rlskoeser (approved):
wow, such a simple fix!

I'm not sure we need it in all these cases... am inclined to only add to `index_pages` where we know it can happen. (But I don't feel strongly about this... trying to think about implications either direction and not sure.)

Looks there are some tests inspecting how this is called, you'll have to update to match. Can be merged as soon as you do!

quadrismegistus:
Up to you! My argument for keeping it would be: in all the instances it's used in a manage.py command to do pretty labor intensive things (generate_corpus, hathi_import and index_pags) and I'm not sure a progressbar's wrong estimate of the total count should be given the power to kill that work with an Exception.

Also, any idea why the javascript unit tests are failing?

rlskoeser:
> .... My argument for keeping it would be: in all the instances it's used in a manage.py command to do pretty labor intensive things (generate_corpus, hathi_import and index_pags) and I'm not sure a progressbar's wrong estimate of the total count should be given the power to kill that work with an Exception.

Good logic! Sold.

> Also, any idea why the javascript unit tests are failing?

I think we've had an occasional error cropping up — looks like it passed when I re-ran it. 🤷

Issue #489, opened 2022-11-16
post-migration cleanup


rlskoeser:
- [ ] remove mysql python dependencies
- [ ] remove mysql from unit test matrix
- [ ] update ansible variables for correct deploy context after renaming unit test matrix

quadrismegistus:
>  update ansible variables for correct deploy context after renaming unit test matrix

So I'm able to leave `deploy_contexts` as is with `'Python unit tests (8.6, postgresql)'` through the solution taken in PR https://github.com/Princeton-CDH/ppa-django/pull/491, since the matrix structure is preserved with the mysql option removed. I thought this preservation would be more modular and congenial to future migrations or testing regimes.

PR #491, opened 2022-11-21
Post migration cleanup

quadrismegistus:
For #489

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -39,18 +39,9 @@ jobs:
     strategy:
       matrix:
         include:
-         - solr: 6.6.5
-           db: mysql
          - solr: 8.6
            db: postgresql
     services:
-      mariadb:
-        image: mariadb
-        env:
-          MYSQL_DATABASE: ${{ env.DB_NAME }}
-          MYSQL_ROOT_PASSWORD: ${{ env.DB_PASSWORD }}
-        ports:
-        - 3306:3306
       postgres:
         image: postgres:12
         env:
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -2,7 +2,7 @@ name: unit tests
 
 on:
   push: # run on every push or PR to any branch
-  # pull_request:  off temporarily
+  pull_request:
   schedule: # run automatically on main branch each Tuesday at 11am
     - cron: "0 16 * * 2"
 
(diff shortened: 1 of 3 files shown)


ci/testsettings.py
```diff
@@ -8,22 +8,16 @@
...
         "HOST": "127.0.0.1",
-        "PORT": "3306",
-        "OPTIONS": {
-            # In each case, we want strict mode on to catch truncation issues
-            "init_command": "SET sql_mode='STRICT_TRANS_TABLES'",
-        },
+        "PORT": "5432",
         "TEST": {
             "CHARSET": "utf8",
-            "COLLATION": "utf8_general_ci",
```
quadrismegistus, line 26: This was unsupported by postgres.

ci/testsettings.py
```diff
@@ -8,22 +8,16 @@
...
+        "ENGINE": "django.db.backends.postgresql",
         "NAME": "ppa",
         "PASSWORD": "ppa",
-        "USER": "root",
+        "USER": "ppa",
         "HOST": "127.0.0.1",
-        "PORT": "3306",
-        "OPTIONS": {
-            # In each case, we want strict mode on to catch truncation issues
-            "init_command": "SET sql_mode='STRICT_TRANS_TABLES'",
```
quadrismegistus, line 22: This was also unsupported by postgres (init_command).

rlskoeser: yes, this is mysql specific

ci/testsettings.py
rlskoeser: right, also mysql specific

requirements.txt
```diff
@@ -1,5 +1,4 @@
 django>=3.2,<4.0
-mysqlclient>=1.3.13
 # avoid a mariadb max field length issue in django-cas-ng>=4.2
 django-cas-ng==4.1.1
```
rlskoeser, line 3: can we unpin this version now, if the problem was mysql specific?

.github/workflows/unit-tests.yml
```diff
@@ -39,18 +39,9 @@ jobs:
     strategy:
       matrix:
         include:
-         - solr: 6.6.5
-           db: mysql
          - solr: 8.6
            db: postgresql
```
rlskoeser, line 43: We should get rid of the matrix entirely since we only want to run one build. And we can go back to using the `.python-version` file in the project to manage the version for the build, which should help with the upgrade. The geniza unit tests have an example of how we're doing that.

rlskoeser (approved):
LGTM!

Issue #490, opened 2022-11-21
Upgrade python from 3.6


quadrismegistus:

rlskoeser:
@quadrismegistus forgot to ask, did you update the python version in ansible and test this upgrade to confirm everything works properly in qa?

quadrismegistus:
Yes— today we did. It meant addressing this issue (https://github.com/Princeton-CDH/cdh-ansible/issues/128) by developing this PR (https://github.com/Princeton-CDH/cdh-ansible/pull/131).

PR #495, opened 2022-11-22
Upgrading python 3.6 to 3.9

quadrismegistus:
In this PR:
- [x] #490
- [x] #493 
- [x] Cleaning some other default settings to make psql default not mysql


rlskoeser (approved):
Great to see that this was so straightforward!

Please look at requirements.txt and see if there were any dependencies we had pinned because we were stuck on python 3.6.  Up to you if you want to handle those as part of this PR or do it in another pass.

rlskoeser:
@quadrismegistus once you merge this, please try updating ansible to this python version and running the QA deploy to confirm that everything works there. The python version is specified in `group_vars/prosody/vars.yml` — changing that _should_ be all that's needed, since the ansible playbook installs and configures the version we have set.

rlskoeser (approved):

PR #505, opened 2022-12-14
Reorganizing solr queries to allow page highlights across work clusters

quadrismegistus:


quadrismegistus:
The data is all here at this point, the solr query can get both page_highlights and work result scores at once. We just need to find a good way to integrate it with the template I think.

.github/workflows/unit-tests.yml
```diff
@@ -29,7 +29,7 @@ jobs:
             ${{ runner.os }}-node-
       - run: npm ci
       - run: npm run test:unit
-      - uses: codecov/codecov-action@v1
+      - uses: codecov/codecov-action@v3
```
quadrismegistus, line 32: Quite the upgrade!

requirements.txt
```diff
@@ -1,7 +1,8 @@
 django>=3.2,<4.0
 django-cas-ng
 pucas>=0.6
-parasolr>=0.8.2
+#parasolr>=0.8.2
+git+https://github.com/Princeton-CDH/parasolr@feature/grouped-response#egg=parasolr  # @NOTE: For testing, allowing page grouping
```
quadrismegistus, line 5: Should we get rid of this? We're gonna merge that PR soon?

rlskoeser: good point. if we're satisfied that the parasolr part is working and stable, we should merge that and update this to install from `develop` (and then we'll release a new version of parasolr before the new ppa release, after the testing — just in case acceptance testing identifies something we missed)

PR #507, opened 2023-01-11
"Multiprocess", a solution for M1 macs, may be making errors on page indexing on QA/prod servers

quadrismegistus:


ppa/archive/management/commands/index_pages.py
```diff
@@ -3,7 +3,10 @@
 """
 
 import queue
-from multiprocess import Process, Queue, cpu_count
+try:
+    from multiprocess import Process, Queue, cpu_count
```
rlskoeser, line 7: maybe a brief comment explaining why this is here, in case we forget?

rlskoeser (approved):
Looks good to me. Might warrant a brief note somewhere about this (dev notes maybe, since this is an issue on some machines and not others?).

rlskoeser:
@quadrismegistus is this still needed? or did we rule it out as not being the problem?

Issue #445, opened 2022-07-27
As a user, I want items with the same title and author to be collapsed automatically so that my search isn’t clouded by repetitive results.


rlskoeser:
## testing notes

first round testing (keyword search not yet working; sorting within clusters known to be not working)

- test clusters imported from unique ids CSV file:
   - [x] clusters should be created as needed from the script
   - [x] cluster creation time/source should be documented in a log entry
   - [x] digitized works should be associated with clusters according to the csv
   - [x] digitized works should have a log entry documenting the change
 - clustering search logic
   - [x] in main archive search, only one work from a cluster is shown
   - [x] when clicking through to search within a cluster, all works are shown
   - [x] cluster icon is not displayed when searching within a cluster
   - [x] when searching within a cluster, search terms and filters are applied within that cluster
- test cluster icon display 
   - [x] works that are part of a cluster include the stacked book icon and text with link to search within the cluster
   - [x] document details page for work within a cluster includes stacked book icon and link to search
   - [x] when searching within the cluster, icon and text indicates at top of page
- review cluster icon styles
   - [x] main search results (desktop/tablet/mobile)
   - [x] document detail page (desktop/tablet/mobile)
   - [x] search within a cluster (desktop/tablet/multiple)

see designs for multiple editions in Zeplin: https://app.zeplin.io/project/5be36b3d9dcaf67e95d71839/screen/60c3e0a8929ba112f9f6ee0f

## dev notes
- [x] add group id to digitized work model
- [x] write migration/script to set edition ids from CSV file
- [x] index group id in solr
- [x] group solr results by edition id so that only the first is shown
- [x] display icon for edition/grouped result in search display
- [x] add link to search within result set 
- [x] handle keyword search; display pages + snippets from the correct volume in the cluster
- [x] #512
- [x] #510
- [x] #519

mnaydan:
**Importation functionality:** I randomly spot-checked a handful of records to make sure clusters were created according to the script with associated digitized works (can't think of a more comprehensive way to test). Looks good. Logging is working - on the main log it appears in stages (first logging the addition of a cluster, followed by the changes to works added to that cluster via csv). This makes sense to me. 

**Clustering search logic:** Can't test all of this yet, but what I did test looks good. A question I have is whether on the main search page if you're just "browsing" (no active search), should the clustered items be collapsed? I feel like they should (maybe that's part of the functionality that isn't working yet?)

[...]

PR #511, opened 2023-01-26
Making sure persistence of search query

quadrismegistus:
Fix to https://github.com/Princeton-CDH/ppa-django/issues/510, itself a small part of https://github.com/Princeton-CDH/ppa-django/issues/445.

### DIFF
--- a/ppa/archive/templates/archive/snippets/cluster_work.html
+++ b/ppa/archive/templates/archive/snippets/cluster_work.html
@@ -1,5 +1,5 @@
 {# NOTE: expects cluster_id to be passed in #}
 <div class="cluster-work icon container" >
     {# TODO: include count if we can easily get it ; search and browse x digitized works within cluster #}
-    <a href="{% url "archive:list" %}?cluster={{ cluster_id }}">search and browse within cluster</a>
+    <a href="{% url "archive:list" %}?cluster={{ cluster_id }}{% if query %}&query={{ query }}{% endif %}">search and browse within cluster</a>
 </div>
--- a/ppa/archive/templates/archive/snippets/cluster_work.html
+++ b/ppa/archive/templates/archive/snippets/cluster_work.html
@@ -1,5 +1,5 @@
 {# NOTE: expects cluster_id to be passed in #}
 <div class="cluster-work icon container" >
     {# TODO: include count if we can easily get it ; search and browse x digitized works within cluster #}
-    <a href="{% url "archive:list" %}?cluster={{ cluster_id }}&?{% querystring_replace page=1 %}">search and browse within cluster</a>
+    <a href="{% url "archive:list" %}?cluster={{ cluster_id }}&{% querystring_replace page=1 %}">search and browse within cluster</a>
 </div>
--- a/ppa/archive/templates/archive/snippets/cluster_work.html
+++ b/ppa/archive/templates/archive/snippets/cluster_work.html
@@ -1,5 +1,5 @@
 {# NOTE: expects cluster_id to be passed in #}
 <div class="cluster-work icon container" >
     {# TODO: include count if we can easily get it ; search and browse x digitized works within cluster #}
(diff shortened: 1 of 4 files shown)


quadrismegistus:
Do we need any other parameters passed? The collections parameters don't seem relevant -- since we're already searching within a specific set of texts? The "sort" parameter? I couldn't find a way to access that via `{{ sort }}` in the template syntax.

ppa/archive/templates/archive/snippets/cluster_work.html
```diff
@@ -1,5 +1,5 @@
 {# NOTE: expects cluster_id to be passed in #}
 <div class="cluster-work icon container" >
     {# TODO: include count if we can easily get it ; search and browse x digitized works within cluster #}
-    <a href="{% url "archive:list" %}?cluster={{ cluster_id }}">search and browse within cluster</a>
+    <a href="{% url "archive:list" %}?cluster={{ cluster_id }}{% if query %}&query={{ query }}{% endif %}">search and browse within cluster</a>
```
rlskoeser, line 4: I think we need to preserve all query parameters (keyword, author, title, date range, even collection).

If you look at the pagination template snippet, we're using a custom template tag called `querystring_replace` that lets us preserve all the search query args but replace the pagination. I think the same thing is relevant here.

rlskoeser, line 4: Here's what I suggest; I think this should preserve all query/form args, but make sure that the cluster search starts on page 1 of the results — you'll want to check, of course.

```suggestion
    <a href="{% url "archive:list" %}?cluster={{ cluster_id }}&?{% querystring_replace page=1 %}">search and browse within cluster</a>
```

rlskoeser (changes requested):
I think we need more args than you're including, but there's a simple way to address.

Do we have any tests to check search results for cluster links? Would be good to have at least one unit test to confirm the behavior.

ppa/archive/templates/archive/snippets/cluster_work.html
quadrismegistus: Hm. I was assuming if we're searching _within_ a cluster, the author, title, date range, and collection would also be the same for each text within the cluster. (Since from Mary's description significant differences in those things would constitute a different cluster). But I suppose we don't need to make that assumption in the code.

quadrismegistus: I see-- nice-- we already have a querystring_replace function which takes a page argument. I just added this and deployed, seems to be working.

ppa/archive/templatetags/ppa_tags.py
```diff
@@ -36,7 +36,7 @@ def querystring_replace(context, **kwargs):
     for key, val in kwargs.items():
         querystring[key] = val
     # return urlencoded query string
-    return querystring.urlencode()
+    return mark_safe(querystring.urlencode())
```
rlskoeser, line 39: any idea why this wasn' t needed before?

rlskoeser (approved):
looks good

ppa/archive/templatetags/ppa_tags.py
quadrismegistus: It gives us `&amp;` in between the query params in the cluster_work.html template instead of `&`.

rlskoeser: yes, that seems like a reasonable use of mark safe!  I just don't understand why we need it now when didn't need it before, since we're using this template tag for the pagination and we don't have that problem there

quadrismegistus: Whew thanks for pointing this out! Glad we don't need it after all.

rlskoeser: me too! thanks for figuring it out

rlskoeser (approved):

Issue #512, opened 2023-01-26
The clusters aren't collapsing on browse anymore; they previously were (see exchange above). By browse I mean when you arrive at the archive page but haven't input any search terms yet.


quadrismegistus:

quadrismegistus:
This is closed with https://github.com/Princeton-CDH/ppa-django/pull/513

PR #513, opened 2023-01-26
Collapse results by cluster even when no keyword present

quadrismegistus:
See https://github.com/Princeton-CDH/ppa-django/issues/512, part of https://github.com/Princeton-CDH/ppa-django/issues/445.

### DIFF
--- a/ppa/archive/models.py
+++ b/ppa/archive/models.py
@@ -446,7 +446,7 @@ class DigitizedWork(ModelIndexable, TrackChangesModel):
     def cluster_id_s(self):
         """
         Convenience function to get a string representation of the cluster (or self if no cluster).
-        Redunadancy 
+        Reduces redunadancy elsewhere.
         """
         return str(self.cluster) if self.cluster else self.index_id()
 
--- a/ppa/archive/models.py
+++ b/ppa/archive/models.py
@@ -443,7 +443,7 @@ class DigitizedWork(ModelIndexable, TrackChangesModel):
         return self.source_id
 
     @property
-    def cluster_id_s(self):
+    def index_cluster_id(self):
         """
         Convenience function to get a string representation of the cluster (or self if no cluster).
         Reduces redunadancy elsewhere.
@@ -796,9 +796,7 @@ class DigitizedWork(ModelIndexable, TrackChangesModel):
             "collections": [collection.name for collection in self.collections.all()]
             if self.collections.exists()
             else [NO_COLLECTION_LABEL],
-            # "cluster_id_s": str(self.cluster) if self.cluster else None,  # @BUG?
-            # "cluster_id_s": str(self.cluster) if self.cluster else index_id,   # moved to magic property function
-            "cluster_id_s": self.cluster_id_s,
+            "cluster_id_s": self.index_cluster_id,
             # public notes field for display on site_name
             "notes": self.public_notes,
             # hard-coded to distinguish from & sort with pages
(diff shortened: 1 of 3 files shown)


ppa/archive/solr.py
```diff
@@ -123,11 +123,10 @@ def query_opts(self):
...
 
-            # if not searching within a group, collapse reprints/editions
-            # @NOTE: This following code may not be necessary?
-            # if not qs_copy.within_cluster_id:
-            # (not expanding, since we only display the first)
-            # qs_copy = qs_copy.filter("{!collapse field=cluster_id_s}")
+            # If not searching within a group, still collapse reprints/editions
+            # Remember that cluster_id_s is now defined as `str(self.cluster) if self.cluster else index_id` in models.py.
+            # So collapsing by "cluster" id implicitly includes works with no cluster id set.
+            qs_copy = qs_copy.filter("{!collapse field=cluster_id_s}")
```
rlskoeser, line 129: Was thinking that this was redundant code, but I see that we bail out early if there are no keyword search terms, since there's a lot of logic we can skip. 

Could we move the `collapse_on` and `collapse_filter` logic currently on lines 160-166 and rely on that for both cases where we collapse? 

Actually, I think it should work to apply the collapse once outside of the if statement, since we always want it and the logic should be the same. Please try that — but if it complicates things too much or causes problems, you can back out to something more like this (but reusing the common variables).

rlskoeser (changes requested):
I think this solution is close, but would like to reduce the redundancy.

Would also be good to revise the unit test for this method to confirm the change.

quadrismegistus:
Weird – the test that broke recently is passing on my local machine. Let's see if it breaks again

ppa/archive/models.py
```diff
@@ -442,6 +442,14 @@ def __str__(self):
             return "%s (%s)" % (self.source_id, self.pages_digital)
         return self.source_id
 
+    @property
+    def cluster_id_s(self):
```
rlskoeser, line 446: I would just name the property `cluster_id` — the `_s` suffix is used for solr dynamic fields, not meaningful here

ppa/archive/solr.py
```diff
@@ -48,7 +48,7 @@ class ArchiveSearchQuerySet(AliasedSolrQuerySet):
         "work_type_s": "work_type",
         "book_journal_s": "book_journal",
         "group_id_s": "group_id",
-        "cluster_id_s": "cluster_id",
+        "cluster_id_s": "cluster_id",    # <- will this conflict with property self.cluster_id_s? should I rename?
```
rlskoeser, line 51: this is for aliasing the solr field name, which uses dynamic field types, so that we can reference it with a meaningful name — I don't think it should conflict with the cluster id property

rlskoeser (approved):
looks ok to me, just a little cleanup on property name and the comment with a question

ppa/archive/models.py
quadrismegistus: Is `.cluster_id` reserved for the numerical django ID for the cluster object, though? That's what seems to be set there on digiworks now.

rlskoeser: is it? I would expect that to be at `.cluster.pk` on the model  — but you've been looking at the code more recently than me!  But maybe a different name would be better — include index somehow to differentiate? `index_cluster_id`

the other `.cluster_id` that you had the commented question on is the object returned by Solr

quadrismegistus: `index_cluster_id` it is! thanks!

rlskoeser (approved):

PR #522, opened 2023-02-15
Index pages custom

quadrismegistus:
Fix for https://github.com/Princeton-CDH/ppa-django/issues/523


rlskoeser (commented):
Logic looks great, and I suspect this will be a really useful addition

Can we add a quick unit test for the new logic?  — Should be able to base it on `TestIndexPagesCommand.test_index_pages` ; give it a source id from the sample digitized works fixture that's already loaded and check that methods are called as expected (e.g. progressbar should get count based on the `page_count` from the db)

rlskoeser (approved):

PR #529, opened 2023-03-07
Quick PR to pairtree fix

quadrismegistus:


ppa/archive/hathi.py
```diff
@@ -308,10 +308,13 @@ def _content_path(self, ext, ptree_client=None):
         # - don't rely on them being returned in the same order on every machine
         parts = pairtree_obj.list_parts(self.content_dir)
         # find the first zipfile in the list (should only be one)
-        filepath = [part for part in parts if part.endswith(ext)][0]
-
-        return os.path.join(pairtree_obj.id_to_dirpath(), self.content_dir, filepath)
-
+        filepaths = [part for part in parts if part.endswith(ext)]
+        return (
```
rlskoeser, line 312: I think we need some kind of output or logging error/warning if this happens, because if the expected file is not found it means something is wrong with the pairtree data — would be good to know which record it's happening on so we can investigate

ppa/archive/models.py
```diff
@@ -1088,32 +1088,34 @@ def hathi_page_index_data(cls, digwork):
...
-                            "content": pagefile.read().decode("utf-8"),
-                            "order": page.order,
-                            "label": page.display_label,
-                            "tags": page.label.split(", ") if page.label else [],
-                            "item_type": "page",
-                        }
-                    except StopIteration:
-                        return
+        zpath=digwork.hathi.zipfile_path()
+        if zpath:
```
rlskoeser, line 1092: same problem here — as written, we won't know that we've failed to find and index page data; an exception in the other method might be enough so we can report the error and move on

ppa/archive/hathi.py
quadrismegistus: I would prefer to log it and not raise an exception because there appear to be a variety of reasons that pairtrees can break down, and the cost of raising an exception is high.

rlskoeser: Was thinking an exception here that we can catch in the indexing script so we can control reporting. Possibly fewer part of the code that need to be touched.

Logging should be ok as long as it's visible when we run the script

quadrismegistus: I see, sorry I didn't see your 2nd comment. yes as long as we're ok with catching it at index function level. I'll do that now

rlskoeser (commented):
Exception logic looks good, now just need to add unit test for the error case.

I'm not sure why, but this PR is showing changes from your solr 8.4 branch (which I thought you merged into develop?)

rlskoeser (approved):
Exception & test case look good, thanks for addressing this.

rlskoeser:
@quadrismegistus would you make a quick bug issue to track this for reporting purposes? Doesn't need too much detail.

Any thoughts on how to test this bug fix in qa? I guess we could manually remove something in the pairtree data and then try to index that record and confirm we get the warning....

PR #532, opened 2023-03-08
Upgrading isort in precommit config

quadrismegistus:
I was getting this error for a while trying to use pre-commit on PPA: https://github.com/home-assistant/core/issues/86892

Upgrading isort here to fix it.


rlskoeser (approved):
Thanks for figuring out the problem & how to fix it!

Issue #556, opened 2023-10-30
Adapt Vineet's script to export plain text corpus


quadrismegistus:

mnaydan:
Developer steps before acceptance testing:
- [ ] Get Jeri shell access to servers
- [ ] Mirror production data to QA
- [ ] Run `index_pages` command to fix blank changed-clusters record bug [(#554)](https://github.com/Princeton-CDH/ppa-django/issues/554)

Acceptance testing checklist:
- [ ] Code runs on QA
- [ ] Any RSE team member can run code
- [ ] Number of json files outputted equals number of public works in PPA (6,752 as of 11/30/23)
- [ ] Suppressed records are being ignored
- **Sample suppressed IDs:**
- nyp.33433069255440
- uc1.$b253881
- dul1.ark:/13960/t07w7ct97
- ien.35556004818043
- CW0106468070
- CB0131406785
- [ ] Both full works and excerpts are contained in corpus
- **Sample full work IDs:**
- uc1.$b14645
- mdp.39015003633594
- wu.89099903650
- loc.ark:/13960/t4rj4z67d
- CW0117319378
- CB0127088549
- **Sample excerpt/article IDs:**
- CW0114589903
- CB0129112818
- coo.31924051399685
- aeu.ark:/13960/t1pg22p71
- uc1.b3924132
- uiuo.ark:/13960/t4qk01n82
- [ ] Both HathiTrust and ECCO works are contained in corpus (should be able to test using IDs above)
- [ ] All expected fields are present for each page (@quadrismegistus provide list)
- [ ] HathiTrust IDs with extra punctuation behave the same as regular IDs
- loc.ark:/13960/t0jt08550
- dul1.ark:/13960/t5j970j9d
- [ ] Works with no pages (e.g., as a result of a bug [(#539)](https://github.com/Princeton-CDH/ppa-django/issues/539)) create a metadata entry but no json file
- uga1.32108002998303

Things to recheck after development:
- [ ] Confirm bugfix to [(#539)](https://github.com/Princeton-CDH/ppa-django/issues/539) resolves the problem in the corpus script output in the way we expect (should pull pages and save a json file)
- [ ] Confirm fix to excerpt range changes in HathiTrust [(#560)](https://github.com/Princeton-CDH/ppa-django/issues/560) resolves the problem in the corpus script output in the way we expect (should pull correct/relevant ranges)

jerielizabeth:
[...]

PR #562, opened 2023-11-28
Generate plain text corpus via export command

quadrismegistus:
For #556

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,6 +5,6 @@ pytest-cov
 django-debug-toolbar
 sphinx
 pre-commit
-wordfreq
-nltk
+pandas
+orjson
 tqdm
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,3 +5,6 @@ pytest-cov
 django-debug-toolbar
 sphinx
 pre-commit
+wordfreq
+nltk
+tqdm
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,6 +5,4 @@ pytest-cov
 django-debug-toolbar
 sphinx
 pre-commit
-pandas
-orjson
 tqdm
--- a/ppa/archive/management/commands/generate_textcorpus.py
+++ b/ppa/archive/management/commands/generate_textcorpus.py
@@ -5,7 +5,7 @@ into Solr via the **index** manage command.
 """
 
 import os
-import orjson
+import orjson   # a faster json implementation
 from django.core.management.base import BaseCommand
 from parasolr.django import SolrQuerySet
 from collections import defaultdict, OrderedDict
--- a/ppa/archive/management/commands/generate_textcorpus.py
+++ b/ppa/archive/management/commands/generate_textcorpus.py
@@ -2,759 +2,187 @@
 **generate_textcorpus** is a custom manage command to generate a plain
 text corpus from Solr.  It should be run *after* content has been indexed
 into Solr via the **index** manage command.
-
-Vineet Bansal authored the parts of the code below that iterate over the Solr-
-indexed corpus. Wouter Haverals authored the functions that preprocess the
-returned page texts: cleaning OCR, rejoining lines, and other useful processes.
-
-By default, *all* documents found in the Solr index are serialized.
-This can be controlled using -`-doc-limit`,
-which denotes the maximum no. of documents to serialize. This is especially
-useful for development, or for sanity-testing your Solr installation.
-
-For corpus generation, the following pre-processing options are available via
-the `--preprocess` flag::
-
-    # Lower-cases words
-    'lower'
-    # Strips HTML tags
-    'strip_tags'
-    # Strips punctuation
-    'strip_punctuation'
-    # Collapses multiple whitespaces into one
-    'strip_multiple_whitespaces'
-    # Strips numeric characters
-    'strip_numeric'
-    # Removes stopwords - Note that the set of default stopwords used by Gensim
-    # is from Stone, Denis, Kwantes (2010).
-    'remove_stopwords'
-    # Strip short words. The lower limit on word length is 3.
-    'strip_short'
-    # Use Porter Stemmer for word-normalization.
-    'stem_text'
-
-IMPORTANT - NO preprocessing filters are applied by default, but you will
-typically at least want to use `lower`.
-Multiple preprocessing filters can be applied (in order) by specifying multiple
-`--preprocess` flags.
-
-Example usage::
-
-    # Save all files to the 'data' folder, with bare-minimum preprocessing
-    python manage.py generate_corpus --path data --preprocess lower
-    --preprocess strip_tags
-
-    # Restrict corpus to 1000 documents
-    python manage.py generate_corpus --path data --doc-limit 1000
-    --preprocess lower --preprocess strip_tags
-
-    # Don't generate dictionary; don't generate metadata
-    python manage.py generate_corpus --path data --doc-limit 1000
-    --preprocess lower --no-dictionary --no-metadata
-
 """
 
-import csv
-import logging
-logging.getLogger().handlers.clear()
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.ERROR)
-
-
-import os.path
-from collections import OrderedDict
-from os import makedirs
-from pprint import pprint,pformat
-
-from django.core.management.base import BaseCommand
-from gensim import corpora
-from gensim.corpora.dictionary import Dictionary
-from parasolr.django import SolrQuerySet
-from progressbar import NullBar, ProgressBar
-
-
-
-### PREPROCESSING CODE ###
-
-PATH_HERE = os.path.abspath(os.path.dirname(__file__))
-PATH_REPO = os.path.abspath(os.path.join(
-    PATH_HERE,
-    '..', # management
-    '..', # archive
-    '..', # ppa
-    '..' # ppa-django
-))
-PATH_REPO_DATA = os.path.join(PATH_REPO, 'data')
-PATH_OCR_RULESETS = os.path.join(PATH_REPO_DATA, 'ocr_cleanup_rulesets')
-
-
-# imports...
-import os,sys,warnings,random
-warnings.filterwarnings('ignore')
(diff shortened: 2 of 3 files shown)


ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+@contextmanager
+def logging_disabled(highest_level=logging.CRITICAL):
+    """Quick way to suppress solr logs as we iterate. Taken from https://gist.github.com/simon-weber/7853144"""
+    previous_level = logging.root.manager.disable
+    logging.disable(highest_level)
+    try: yield
+    finally: logging.disable(previous_level)
+
```
rlskoeser, line 174: This should be handled by updating django logging config rather than overriding here (we might need to turn logging on to debug something). What log level are you using and how noisy is it ?

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,200 @@
...
+        # save docs and gather metadata
+        metadata=[]
+        pdesc='Saved text to'
+        pbar=tqdm(total=self.doc_count, desc=f'{pdesc}: ...')
+        for doc_id in self.doc_ids:
+            # get saved filename and found metadata for this document
+            fn,meta = self._save_doc(doc_id)
+
+            # if we saved, update progress bar desc
+            if fn: pbar.set_description(f'{pdesc}: {fn}')
```
rlskoeser, line 149: this doesn't look like standard formatting, do you have the pre-commit hooks installed?

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,200 @@
...
+        results = SolrQuerySet().facet(self.DOC_ID_FIELD, limit=self.doc_limit)
+        # store page counts and doc ids
+        self.page_counts = results.get_facets().facet_fields[self.DOC_ID_FIELD]
+        self.doc_ids = self.page_counts.keys()
+        self.doc_count = len(self.doc_ids)
+    
+    @staticmethod
+    def _get_id(doc_id:str) -> str:
+        """Method to make a file-safe version of a document ID"""
+        return doc_id.replace('/','|')
```
rlskoeser, line 63: django may have an existing escape character utility method we can use that handles more characters

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # limit docs queried
+        self.doc_limit = doc_limit
+
+        # subsequent paths
+        self.path_texts = os.path.join(self.path,'texts')
+        self.path_metadata = os.path.join(self.path,'metadata.json')
+        
+        # query to get initial results
+        results = SolrQuerySet().facet(self.DOC_ID_FIELD, limit=self.doc_limit)
```
rlskoeser, line 54: why use facets?

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,200 @@
...
+        meta = {self.WORK_ID_FIELD:work_id, **metadata_docs[0]}
+
+        # find the pages docs
+        page_docs = [d for d in docs if d["item_type"] == "page"]
+
+        # transform into new dictionary with keys in `self.PAGE_ID_FIELD` and `self.OUTPUT_DOC_FIELDS`
+        pages = [self._transform_doc(doc,meta) for doc in page_docs]
+
+        # make sure sorted by numeric page num (i.e. "digital")
+        pages.sort(key=lambda page: page[self.PAGE_SORT_FIELD])
```
rlskoeser, line 97: Is the Solr sorting unreliable?

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,200 @@
...
+            SolrQuerySet()
+            .search(**{self.DOC_ID_FIELD: doc_id})
+            .order_by(self.PAGE_ORDER_FIELD)
+        )
+
+        # populate the result cache with number of rows specified
+        docs = [
+            doc
+            for doc in result.get_results(rows=self.page_counts[doc_id])
+            if doc[self.DOC_ID_FIELD]==doc_id
```
rlskoeser, line 82: is the Solr filter unreliable?

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,200 @@
...
+        ]
+
+        # find the metadata doc
+        metadata_docs = [d for d in docs if d["item_type"] == "work"]
+        assert len(metadata_docs)==1
+        meta = {self.WORK_ID_FIELD:work_id, **metadata_docs[0]}
+
+        # find the pages docs
+        page_docs = [d for d in docs if d["item_type"] == "page"]
```
rlskoeser, line 91: I'm having trouble following, why is this part needed?

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,200 @@
...
+            for key_new,key_orig in (
+                self.OUTPUT_DOC_FIELDS.items()
+            )
+        }
+
+        # return with page id
+        return {
+            self.PAGE_ID_FIELD:f'{meta[self.WORK_ID_FIELD]}_{odoc[self.PAGE_NUM_FIELD]}',
+            **odoc
+        }
```
rlskoeser, line 115: This is hard to understand. What's going on, and why is it needed? The docstring should provide more context and an example if that's needed to make this clear.

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,200 @@
...
+from django.core.management.base import BaseCommand
+from parasolr.django import SolrQuerySet
+from collections import defaultdict, OrderedDict
+import logging
+from tqdm import tqdm
+from typing import Tuple
+from contextlib import contextmanager
+import logging
+
+class SolrCorpus:
```
rlskoeser, line 18: I vaguely remember this class from Vineet's code. It's great to keep it if you find it a useful way to encapsulate the logic. (I would probably put the methods on the manage command, personally.)

If you keep it, you should write unit tests for each method of the class.

ppa/archive/tests/test_generate_textcorpus.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_invalid_preprocess_flags(tmpdir, patched_solr_queryset):
+    # Flags that are not supported
+    with pytest.raises(CommandError):
+        call_command(
+            "generate_textcorpus", "--path", tmpdir.dirpath(), "--doc-limit","one"
+        )
+
+    with pytest.raises(CommandError):
```
rlskoeser, line 100: fyi, `pytest.raises` has optional arguments that will match the expected error text in the exception

ppa/archive/management/commands/generate_textcorpus.py
rlskoeser: I think a regular search would be better than a facet. You need to add a filter on suppressed/public here. It could be helpful to use our existing ppa solr queryset object here.

dev-requirements.txt
```diff
@@ -4,4 +4,5 @@ pytest-django>=4.5.2
 pytest-cov
 django-debug-toolbar
 sphinx
-pre-commit
\ No newline at end of file
+pre-commit
+tqdm
```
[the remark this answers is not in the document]
rlskoeser: while you're at it: could you check if we're still requiring `gensim` and comment it out? we should make it an optional install since we're not planning to run the other version of this script

rlskoeser (commented):
It's really great to see a draft of this script. It looks like you were able to adapt some of Vineet's code; glad to see that. (although I haven't compared so it's hard to tell what you've changed. I do remember his version is far more complicated than we needed!)

I'm having trouble understanding some of the transformation logic - both what you're doing but more importantly why you're doing it. You should be able to get what you need out of Solr pretty directly with minimal transformation in the python code. If you need to rename the fields, you can do that when you limit which fields are returned - you can read about it in the Solr docs or look at our solr queryset object where we're doing something similar (we call it `aliases` there).

I'm also a little surprised about the number of solr queries - is it one query to get the documents and then one query per document to get the pages? Maybe this is the same as what Vineet was already doing, I don't remember, but I think we could do it more efficiently.  I'd be happy to talk through different approaches. I guess right now you're writing out one file of page content per document so maybe this is reasonable and we can just keep it in mind as a possibility for future refinement.

rlskoeser:
@quadrismegistus have you used json lines before (JSONL), and have you considered using it for the PPA corpus?

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: Yes, I found it was in one instance. I can try to recreate

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: We're just returning a new output dictionary with field names changed and an additional field called page_id which is equal to {work_id}_{page_orig}.

quadrismegistus:
> @quadrismegistus have you used json lines before (JSONL), and have you considered using it for the PPA corpus?

Yes. What benefit do you see it having in this instance? Are you imagining a .jsonl file for the entire corpus, or changing the individual work jsons to jsonls? The former seems unwieldy to me for average users and the latter seems unnecessary given that the json files are small, read in <1ms, and the jsonl file would eventually read the same file contents into memory and I'm not sure there'd by any speed improvement.

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: I found it was in one instance, yes, I can try to recreate.

quadrismegistus:
@rlskoeser I started a new branch/PR for a minimalist jsonl implementation of this, branching off this PR, because it kind of gets rid of a lot, but I kind of like the simplicity now: #563

I'll request review there and if you like that approach I'll merge that PR into this one.

PR #563, opened 2023-12-04
Generate text corpus as jsonl

quadrismegistus:
An alternative approach to text corpus generation: two files:
* metadata.json
* pages.jsonl.gz


ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -191,10 +28,36 @@ def add_arguments(self, parser):
...
+        i=0
+        q=SolrQuerySet().search(item_type=item_type)
+        total = q.count()
+        for i in progressbar(range(0, total, nsize)):
+            q.set_limits(i,i+nsize)
+            yield from q
+
+    def iter_pages(self):
+        for d in self.iter_solr(item_type='page'):
+            yield {k:v for k,v in d.items() if k in self.PAGE_OUTPUT_FIELDS}
```
rlskoeser, line 41: let Solr do this work for you - in parasolr queryset the method is `only` (inspired by django queryset) https://parasolr.readthedocs.io/en/latest/codedocs.html#parasolr.query.queryset.SolrQuerySet.only

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -191,10 +28,36 @@ def add_arguments(self, parser):
             "The default of -1 considers ALL documents.",
         )
 
+    def iter_solr(self, nsize=10, item_type='page'):
+        i=0
+        q=SolrQuerySet().search(item_type=item_type)
+        total = q.count()
+        for i in progressbar(range(0, total, nsize)):
+            q.set_limits(i,i+nsize)
```
rlskoeser, line 36: why do you need to set limits every iteration? should this be an update call? (could be I've forgotten how to use progressbar, it's been a while)

rlskoeser: while I'm thinking about it: would be good to make progressbar optional, since it's possibly we'll want to run this script without that (e.g. if we set it up as a cron job)

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -191,10 +28,36 @@ def add_arguments(self, parser):
...
+        path = options['path']
+        os.makedirs(path, exist_ok=True)
+        path_meta = os.path.join(path,'metadata.jsonl')
+        path_texts = os.path.join(path,'pages.jsonl')
+        
+        with jsonlines.open(path_meta,'w') as of_meta:
+            for i,d in enumerate(self.iter_works()):
+                of_meta.write(d)
+                if options['doc_limit']>0 and i+1>=options['doc_limit']: 
+                    break
```
rlskoeser, line 57: I missed that you were generating the metadata export in this script as well. Would like to align with the admin export (minus internal fields, but I guess that can be a later step.

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -191,10 +28,36 @@ def add_arguments(self, parser):
...
+        
+        with jsonlines.open(path_texts,'w') as of_meta:
+            for i,d in enumerate(self.iter_pages()):
+                of_meta.write(d)
+                if options['doc_limit']>0 and i+1>=options['doc_limit']: 
+                    break
```
rlskoeser, line 63: why is the enumeration and doc limit filtering needed?

would likely be more efficient to iterate and save in chunks

ppa/archive/management/commands/generate_textcorpus.py
rlskoeser: I don't think the metadata is large enough to merit jsonl and don't think it needs to be the same

rlskoeser (commented):
Thanks for the quick spike on this - I'd be interested in taking the output for a spin; I think jsonl might be a good option for us, more efficient to generate and work with.  

How do we make the call / who should make the call?

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: the `set_limits` is from parasolor as method of the SolrQuerySet. progressbar ended up being simple and nearly identical to tqdm. will add an verbose/quiet flag

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: Is the solr and admin metadata (or another export command) guaranteed to be in sync?

Agreed, metadata can be json easily, will change back.

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: jsonlines by default saves by row, I'm not aware of chunking. The doclimit param is a little silly, legacy of previous script, I don't see why anyone would want a partial export where for the gensim script it made more sense, maybe we can just can it.

rlskoeser: If we need a doclimit, I think it should be applied at the solr query level - maybe a page limit would make more sense for the new version.  How long does the script take? Any value in a shorter run for dev/test? If not I agree about dropping the option.

ppa/archive/tests/test_generate_textcorpus.py
```diff
@@ -55,35 +55,26 @@ def patched_solr_queryset(mock_solr_queryset):
         "ppa.archive.management.commands.generate_textcorpus.SolrQuerySet", new=mock_qs
     ) as mock_queryset_cls:
         mock_qs = mock_queryset_cls.return_value
-        mock_qs.get_results.return_value = mock_solr_docs
-        mock_qs.get_facets.return_value.facet_fields = mock_solr_facets
-
+        mock_qs.only.return_value.count.return_value = len(mock_solr_docs)
+        mock_qs.only.return_value.get_results.return_value = mock_solr_docs
```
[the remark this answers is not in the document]
rlskoeser: yes! we have a pytest plugin: https://github.com/Princeton-CDH/parasolr/blob/main/parasolr/pytest_plugin.py#L160

maybe it isn't included in the docs yet? (if you can confirm missing please create an issue or fix if it's easy; please note/suggest any other doc improvements, your perspective would be helpful)

ppa/archive/management/commands/generate_textcorpus.py
rlskoeser: ah, I forgot about that syntax because I never use it directly. SolrQuerySet is designed to support indexing like a list, so you should be able to do `[100:200]` to get a chunk of results and it should query Solr for you.  

quite possible this is not documented or not clearly documented (please flag/help resolve)

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -2,199 +2,238 @@
...
-    """Quick way to suppress solr logs as we iterate. Taken from https://gist.github.com/simon-weber/7853144"""
-    previous_level = logging.root.manager.disable
-    logging.disable(highest_level)
-    try: yield
-    finally: logging.disable(previous_level)
+        # define a generator to iterate solr with
+        iterr = (
+            result
+            for step in range(0, total, batch_size)
+            for result in qset[step : step + batch_size]
```
rlskoeser, line 144: cool

rlskoeser (commented):
Started testing but discovered I didn't have page data in my dev solr collection.   I had some problems running it but I think they all are due to zero page results from Solr.  Successfully running it now with partial page data.

Preliminary comments:
- could we shorten the name of the manage script? 
- need to add orjsonl to requirements now that we've decided to go with that
- script crashes when there are no pages in solr; I know we are relying on the indexing but quick sanity check on page content would be helpful. Ideal would be page count in solr matches (or is close to) the total number of pages reported in the postgres database (now I'm curious how closely those two things match....). 

Looks like everything is working well, I'll reindex pages locally overnight and test a little tomorrow and let you know if I have any more feedback.

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -2,199 +2,238 @@
...
-        pbar.close()
+        # get params
+        batch_size, lim = self.batch_size, self.doclimit
 
-        # save metadata csv
-        with open(self.path_metadata,'w') as of:
-            json.dump(metadata, of, indent=2)
-        print(f'Saved metadata to: {self.path_metadata}')
+        # search for this item type, order by id
+        qset = self.query_set.search(item_type=item_type).order_by("id")
```
rlskoeser, line 119: I think `filter` might be slightly better than `search` based on how Solr caches things, although it might not make a difference here.

```suggestion
        qset = self.query_set.filter(item_type=item_type).order_by("id")
```

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -2,199 +2,238 @@
...
+        yield from iterr
 
+    def iter_works(self):
+        """
+        Simply calls `.iter_solr()` with `item_type` as `'work'`
+        """
+        for d in self.iter_solr(item_type="work"):
+            # source does not need to be a list, will only be one
+            assert len(d["source"]) == 1
+            d["source"] = d["source"][0]
```
rlskoeser, line 161: I had been thinking of using the aliased solr queryset (which lets you rename fields)  in parasolr, which now has a method for updating the result document before it returns. I kind of suspect your generator solution is more efficient, though.

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -2,199 +2,238 @@
...
+        # yield from this generator, progress bar or no
+        yield from iterr
 
+    def iter_works(self):
+        """
+        Simply calls `.iter_solr()` with `item_type` as `'work'`
+        """
+        for d in self.iter_solr(item_type="work"):
+            # source does not need to be a list, will only be one
+            assert len(d["source"]) == 1
```
rlskoeser, line 160: I don't think this assertion is needed; if it does fail, the whole script will abort - but it also shouldn't fail! Can we make the assumptions clearer in the higher level comments or docstring?
```suggestion
```

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -2,199 +2,238 @@
...
+    def save_pages(self):
+        """
+        Save the page-level metadata as a jsonl file
+        """
+        ### save pages
+        if self.is_dry_run:
+            # iterate in place
+            deque(self.iter_pages(), maxlen=0)
+        else:
+            # save to jsonl or jsonl.gz
```
rlskoeser, line 204: What controls whether it is gzipped or not? is it just based on the filepath?

If it's easy, would be nice to have a flag to generate plain jsonl - seems useful for dev at least.

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -2,199 +2,238 @@
...
 
-@contextmanager
-def logging_disabled(highest_level=logging.CRITICAL):
-    """Quick way to suppress solr logs as we iterate. Taken from https://gist.github.com/simon-weber/7853144"""
-    previous_level = logging.root.manager.disable
-    logging.disable(highest_level)
-    try: yield
-    finally: logging.disable(previous_level)
+        # define a generator to iterate solr with
+        iterr = (
```
rlskoeser, line 141: can we give this a more semantic name? would `chunks` or `chunk_iterator` make sense here?

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -2,199 +2,238 @@
...
+DEFAULT_BATCH_SIZE = 10000
+
+
+class Command(BaseCommand):
+    """
+    Custom manage command to generate a text corpus from text indexed in Solr.
+    """
+
+    # fields we want from pages, in solr syntax: newfieldname:oldfieldname
+    PAGE_FIELDLIST = [
```
rlskoeser, line 37: any reason you didn't use a dict? (that's what we use in the aliased queryset in parasolr)

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -50,12 +50,16 @@ class Command(BaseCommand):
...
         "pub_place:pub_place",
         "collections:collections_exact",
         "work_type:work_type_s",
-        "sources:source_t",
+        "source:source_t",
         "source_url:source_url",
+        # title_sort for better key sorting, but could be sort_title
+        "title_sort:sort_title",
+        # title_sub for better key sorting, but could be subtitle
+        "title_sub:subtitle",
```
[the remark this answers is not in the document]
rlskoeser: I see the `sort_keys` option on the `json.dump` call now, new to me (and glad to know about it!). What order do we get without that?

rlskoeser (commented):
@quadrismegistus this is fantastic work. I have a few small suggestions for the existing code, but the approach looks great. From a code perspective jsonl seems like a good solution.

From the debug logging it looks like parasolr is hitting Solr twice for every chunk of results; I suspect that is a parasolr problem because I don't see anything in your code that seems like it would cause that. 

I'm getting an error from Solr when I try to export all pages, have you seen anything like this?  (Error is parasolr request, not the new code)
```
requests.exceptions.JSONDecodeError: Invalid control character at: line 55045 column 5572 (char 20851707)
```

Please update this PR to point to develop when you get a chance.

LMK what you need from me on the unit tests so we can get this merged and delivered for acceptance testing.

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -2,199 +2,238 @@
...
 
-            # if we saved, update progress bar desc
-            if fn: pbar.set_description(f'{pdesc}: {fn}')
+    @cached_property
+    def query_set(self):
+        """
+        A cached representation of the query set;
+        it discourages opening multiple connections to database.
+        """
+        return SolrQuerySet()
```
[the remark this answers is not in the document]
quadrismegistus: yep, that would work just as well

manage.py
```diff
@@ -11,7 +11,7 @@
         # issue is really that Django is missing to avoid masking other
         # exceptions on Python 2.
         try:
-            import django
+            pass
```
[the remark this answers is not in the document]
quadrismegistus: woops -- I didn't mean to commit any changes to manage.py

ppa/archive/management/commands/generate_textcorpus.py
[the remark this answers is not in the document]
quadrismegistus: Could be by dictionary key entry time/order if on a modern Python3 branch. It's not a big issue for our current work, just a practice I generally maintain. But you're right, consistency trumps that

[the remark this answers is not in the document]
rlskoeser: I appreciate the attention to the order of the output, I think that's valuable! Maybe there's a different way to manage that besides alpha key

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -2,199 +2,238 @@
...
-            json.dump(metadata, of, indent=2)
-        print(f'Saved metadata to: {self.path_metadata}')
+        # search for this item type, order by id
+        qset = self.query_set.search(item_type=item_type).order_by("id")
 
+        # if we're looking for pages, rename those fields
+        if item_type == "page":
+            qset = qset.only(*self.PAGE_FIELDLIST)
+        elif item_type == "work":
+            qset = qset.only(*self.WORK_FIELDLIST)
```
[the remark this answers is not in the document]
quadrismegistus: yes! I've been meaning to implement this. what is the field to query on there? I didn't see it in the solr original fields:

```
 "_version_": 1757837811292045312,
   "author": "Sheridan, Thomas, 1719-1788",
   "author_exact": "Sheridan, Thomas, 1719-1788",
   "cluster_id_s": "sheridancompleteSV",
   "collections": [
     "Dictionaries",
     "Linguistic"
   ],
   "collections_exact": [
     "Dictionaries",
     "Linguistic"
   ],
   "group_id_s": "CB0126082904",
   "id": "CB0126082904",
   "item_type": "work",
   "last_modified": "2023-02-14T19:59:33.011Z",
   "order": 0,
   "pub_date": 1796,
   "pub_place": "Philadelphia ",
   "publisher": "W. Young, Mills & Son, booksellers and stationers, no. 52, Second-Street, corner of Chesnut-Street",
   "sort_title": "complete dictionary of the English language, both with regard to sound and meaning: one main object of which is, to establish a plain and permanent standard of pronunciation. To which is prefixed, a prosodial grammar. By Thomas Sheridan, A.M.",
   "source_id": "CB0126082904",
   "source_t": [
     "Gale"
   ],
   "source_url": "https://link.gale.com/apps/doc/CB0126082904/ECCO?u=prin77918&sid=gale_api&xid=a5b78388",
   "subtitle": "both with regard to sound and meaning: one main object of which is, to establish a plain and permanent standard of pronunciation. To which is prefixed, a prosodial grammar. By Thomas Sheridan, A.M.",
   "title": "A complete dictionary of the English language,",
   "work_type_s": "full-work"
```

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: Sure. fwiw this works ok, so at least we know it's a good assumption for the current solr indexing logic. we could add some notes about what this script expects to exist and in what form when we pull from solr, and just pull `d["source"][0]` if we want

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: Yes, just based on filename— we could either add a `--pages-filename` argument (right now we only have an output directory flag) or a `--uncompressed` argument?

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: oh ok, didn't realize— I thought parasolr wanted a list.

ppa/archive/management/commands/generate_textcorpus.py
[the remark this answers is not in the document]
rlskoeser: wow, the danger of trusting my memory. I just checked, you don't need to filter in solr because it isn't there! when a work is suppressed we index a stub record (id only) to make sure we replace any existing data. Sorry for misleading!

I think this does make the check on totals more important, though (# unsuppressed works, # pages expected in the db, actuals in the Solr index). I think I've done something like that elsewhere and can dig it up so we can adapt.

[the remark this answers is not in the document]
quadrismegistus: Just confirming: neither suppressed works nor the pages of suppressed works are in solr?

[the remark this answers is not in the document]
rlskoeser: Correct, that is what we expect (as long as indexing is in sync with the db). Want to make a note in the comments in the code as a reminder?

ppa/archive/management/commands/generate_textcorpus.py
rlskoeser: Makes more sense to me to make the uncompressed a flag vs filename, I think we want that to stay the same. Shorter is better, what about `--no-zip` or similar?

ppa/archive/management/commands/generate_textcorpus.py
rlskoeser: you're generating the correct syntax! parasolr uses the dict to generate that syntax and a few other things: https://github.com/Princeton-CDH/parasolr/blob/main/parasolr/query/aliased_queryset.py#L20-L28

aliased queryset isn't in the parasolr docs either, is it? 🤦‍♀️ 😞

rlskoeser:
@quadrismegistus for reporting on / comparing total number of unsuppressed works and pages in solr and db, we have some similar code in the index pages manage command:

https://github.com/Princeton-CDH/ppa-django/blob/main/ppa/archive/management/commands/index_pages.py#L97-L98

`DigitizedWork.items_to_index` returns a django queryset of items that should be indexed, so you can call `.count()` on that for a total. `Page.total_to_index()` runs an aggregation query to sum all the page counts on the items to be indexed.

At the end of the index_pages script we report on the totals of items by type in solr:
https://github.com/Princeton-CDH/ppa-django/blob/main/ppa/archive/management/commands/index_pages.py#L122-L132

At the beginning of the corpus export script, it would be great to output the totals for both and warn loudly if they don't match. They _should_ match but we'll have to see whether or not they do actually match in practice - I just checked and my page counts are somewhat different in local dev, but that could be a db/pairtree mismatch.

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,262 @@
+"""
+**generate_textcorpus** is a custom manage command to generate a plain
+text corpus from Solr.  It should be run *after* content has been indexed
+into Solr via the **index** manage command.
+
+Assumptions:
+    * No suppressed works from PPA exist in the solr index
+    * The "source_t" field exists in solr index, is a list, 
+        but contains only entry (it will be ["Gale"] or ["Hathi"])
+
```
rlskoeser, line 10: suggested revision

```suggestion
The full text corpus is generated from Solr; this script makes the following assumptions:
    * Suppressed works and pages are not indexed (expected behavior, if db and Solr are synchronized)
    * Each work has a single source (Gale or Hathi) even though it is indexed in Solr as a list
    
```

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,262 @@
...
+        # if we're looking for pages, rename those fields
+        if item_type == "page":
+            qset = qset.only(**self.PAGE_FIELDLIST)
+        elif item_type == "work":
+            qset = qset.only(**self.WORK_FIELDLIST)
+
+        # get the total count for this query
+        total = qset.count()
+        if not total:
+            raise Exception("no records found in solr for query")
```
rlskoeser, line 138: this should raise a django `CommandError`

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,262 @@
...
+        # (this has the effect of pulling from solr only what we need,
+        # since we limit how many rows we pull by this amount)
+        if lim and total > lim:
+            total = lim
+        # define a generator to iterate solr with
+        batch_iterator = (
+            result
+            for step in range(0, total, batch_size)
+            for result in qset[
+                step : (step + batch_size) if (step + batch_size) <= total else total
```
rlskoeser, line 150: was this causing an error?

ppa/archive/tests/test_generate_textcorpus.py
```diff
@@ -0,0 +1,495 @@
...
+        page_iter = cmd.iter_pages()
+        assert isinstance(page_iter, types.GeneratorType)
+
+        res = list(page_iter)
+        assert res, "No pages returned"
+        assert len(res) == len(mock_page_docs)
+        mock_iter_solr.assert_called_with(item_type="page")
+
+
+# def test_iter_solr_lim_works():
```
rlskoeser, line 216: Don't forget to clean these up. Have these been replaced?

ppa/archive/tests/test_generate_textcorpus.py
```diff
@@ -0,0 +1,495 @@
...
+        cmd = init_cmd()
+        assert cmd.path == "ppa_corpus_" + generate_textcorpus.nowstr()
+        assert not cmd.doclimit
+        assert not cmd.verbose
+        assert not cmd.uncompressed
+        assert cmd.batch_size == generate_textcorpus.DEFAULT_BATCH_SIZE
+    os.chdir(herenow)
+
+
+## cannot get this to work ?
```
rlskoeser, line 378: is this still a problem?

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,262 @@
...
+            ]
+        )
+
+        # if progress bar wanted, tack one on
+        if self.progress:
+            batch_iterator = progressbar(batch_iterator, max_value=total)
```
rlskoeser, line 156: to test this, just patch `progressbar` with a mock and you can confirm that is / is not called as expected based on the configuration

ppa/archive/management/commands/generate_textcorpus.py
```diff
@@ -0,0 +1,262 @@
+"""
+**generate_textcorpus** is a custom manage command to generate a plain
```
rlskoeser, line 2: please add this to the sphinx docs - you can put it near where the other manage commands are listed

rlskoeser (commented):
Revisions look good, looks like there is just some minor cleanup left.

Is there anything on our list that hasn't been addressed? I don't see any warning or summary counts, just the new error if the total from solr is zero. This is such a massive PR I'm inclined to get it merged sooner than later - any fine-tuning things that are left  could be done as a smaller PR once this is merged.

rlskoeser (approved):
Sorry, should have said - you have my ok to merge this once you deal with the remaining cleanup items in this PR, I don't need to review again

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: yes! 😬 We needed to make sure the batch_size doesn't push us over the limit of what we want to pull from solr.

rlskoeser: is that a parasolr flaw?

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: should I remove `generate_corpus` or add `generate_textcorpus` beside it?

rlskoeser: Is that Vineet's script? My inclination is to remove it. Can you imagine any scenario where we'd want to run that now that we have the new version?

ppa/archive/management/commands/generate_textcorpus.py
quadrismegistus: not really, since solr doesn't know otherwise what our requested limit is. Just a mistake in how I thought resetting the batch_size might work, which is what I did earlier

Issue #565, opened 2023-12-19
add an option to make page indexing more efficient


rlskoeser:
we often need to reindex all page content in ppa, but when we hit an error that causes the script to crash we have to reindex everything

would be helpful for developer sanity to have an opt-in / non-default way to index only the content that needs it - maybe we can determine by comparing page counts in the database and solr? or maybe we can use modification time to figure out which works need pages indexed? or some combination of the two?

quadrismegistus:
Would this general pattern work?

1. Get a map of {work_id} to its [{page_id}, {page_id}, ...] from django db. That gives us # pages per work too.
2. Query for # of pages per work_id.
3. For those with mismatches, reindex

rlskoeser:
We can easily get a count of total pages / total works; should be able to do a facet query in solr to get number of page items per source id (group id?). Could filter by modification time as well so we only count recently indexed pages and index works that don't have enough recent pages

rlskoeser:
Was able to use this script in qa to do partial reindex after the full page index crashed part way through. We still need to figure out why there's a discrepancy in page counts between db and solr, but we're tracking that on #567

PR #569, opened 2023-12-20
Add option to expedite page indexing based on page count mismatches

rlskoeser:
meant to address #565

started on this when we were fighting with the page indexing; polished a little enough to push for review and be readable

LMK what you think of the approach (whenever)


ppa/archive/management/commands/index_pages.py
```diff
@@ -100,6 +105,55 @@ def handle(self, *args, **kwargs):
             digiworks = DigitizedWork.objects.filter(source_id__in=source_ids)
             num_pages = sum([dw.page_count for dw in digiworks])
 
+        # if reindexing everything, check db totals against solr
+        if not source_ids and self.verbosity >= self.v_normal:
+            # check totals
+            solr_count = self.get_solr_totals()
+
+            work_diff = digiworks.count() - solr_count.get("work", 0)
+            page_diff = num_pages - solr_count.get("page", 0)
```
quadrismegistus, line 114: Is this `digiworks.count()` (the total pages in psql db?) guaranteed to be the sum of the individual work page counts that are checked?

quadrismegistus (approved):
LGTM! Thanks for doing this.

ppa/archive/management/commands/index_pages.py
rlskoeser: `digiworks.count()` is the number of digitized works in the db that we expect to be indexed (i.e. public/non-suppressed)

`num_pages` when calculated from `Page.total_to_index` is a sum of all the page counts in the db for published works. When I was testing locally I saw some discrepancies between these numbers, and I need to review the code to look at how that page total is calculated and why it could be different than what's indexed in Solr. (It seems like it must be a bug if they don't match, which is entirely possible.)

I think even if the page counts don't match exactly, this option still gives us a shortcut to substantially decrease the number of pages we index and is still useful even without/before resolving any page count discrepancies between solr and db.

rlskoeser:
@quadrismegistus thanks for reviewing

Issue #567, opened 2023-12-20
Discrepancy between total works / pages as reported by database and solr


jerielizabeth:
Notes pulled from our testing notes:

Checked totals via django console.

Database reports: 6751 public works, **1938565** pages (based on digitized work page count)
Solr reports: 6751 works,  **1950698** pages

>>> from ppa.archive.models import DigitizedWork, Page
>>> DigitizedWork.objects.filter(status=DigitizedWork.PUBLIC).count()
6751
>>> Page.total_to_index()
1938565
>>> from parasolr.django import SolrQuerySet
>>> SolrQuerySet().all().facet("item_type").get_facets().facet_fields.item_type
OrderedDict([('page', 1950698), ('work', 6751)])

Export is based on Solr, so we expect it to match the solr count
% wc ppa_pages.jsonl
 1950698 597452820 4145091420 ppa_pages.jsonl

Discrepancy between database page count and solr page count is a concern but separate from testing the corpus export script, which is pulling page content from solr as expected 

## testing notes
- [x] check the admin log entries for works with updated page counts; are they present? do they provide the right information?
- [x] try editing an excerpt and changing the digital page range; confirm the page count is updated when you save

(currently based on notes in the code we don't expect page count to update if you remove a page range; we could revisit that in future if it's a possible scenario)

rlskoeser:
As part of testing the new option for the page indexing script (#569 )  I have a list of db/solr page count mismatches from QA. Hopefully a helpful starting point for investigation.  It looks like these are all Hathi ids, so there must be some discrepancy between how we count pages for the db vs how we actually get pages for indexing.  

[ppa-pagecount-mismatches.txt](https://github.com/Princeton-CDH/ppa-django/files/14145804/ppa-pagecount-mismatches.txt)

rlskoeser:
I did a little spot-checking based on this list, and what I found for every record I tried is a discrepancy between the page count value stored in the database and the number I get when I recalculate the page count based on the hathitrust data. I'm wondering if this is another place where we're not accounting for updates to the hathitrust data and so things are getting out of sync.

I'm going to create a quick utility script we can run to update the page counts in the database, but we should keep this mind along with the other related items (rsync, excerpt page ranges) - maybe we can consolidate the updates somehow.

rlskoeser:
[...]

PR #594, opened 2024-02-29
Manage command to update hathitrust page counts

rlskoeser:
Adds a new manage command to update page counts for non-excerpted HathiTrust works.

Reports on what was done, and creates log entries so an admin will be able to see the change.

I haven't created any unit tests since I'm not sure we will actually want to keep the script in this form - it seems that there may be multiple things that we'll want to do when we updated hathitrust content (rsync, check excerpts, recalculate pages...)

related to #567 

Currently the `count_pages` method saves the model if changed (unless it's an excerpt!); based on the difficulties I had tripping over that here and some other related investigations, I'm planning to refactor that so it does not save and make sure the calling code handles saving when appropriate - but this can be reviewed independent of that refactor.

### DIFF
--- a/ppa/archive/management/commands/update_hathi_pagecounts.py
+++ b/ppa/archive/management/commands/update_hathi_pagecounts.py
@@ -55,7 +55,8 @@ class Command(BaseCommand):
                 # recalculate page count from pairtree data
                 # NOTE: this method automatically saves if page count changes
                 digwork.page_count = digwork.count_pages()
-                if old_page_count != digwork.page_count:
+                if digwork.has_changed("page_count"):
+                    digwork.save()
                     stats["updated"] += 1
                     # create a log entry documenting page count change
                     LogEntry.objects.log_action(
--- a/ppa/archive/management/commands/update_hathi_pagecounts.py
+++ b/ppa/archive/management/commands/update_hathi_pagecounts.py
@@ -79,7 +79,7 @@ class Command(BaseCommand):
                 stats["missing_data"] += 1
 
         # report a summary of what was done
-        if verbosity >= self.v_normal:
+        if self.verbosity >= self.v_normal:
             self.stdout.write(
                 f"""Volumes with updated page count: {stats['updated']:,}
     Page count unchanged: {stats['unchanged']:,}
--- a/ppa/archive/management/commands/update_hathi_pagecounts.py
+++ b/ppa/archive/management/commands/update_hathi_pagecounts.py
@@ -1,5 +1,3 @@
-from collections import defaultdict
-
 from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.contrib.admin.models import CHANGE, LogEntry
@@ -20,6 +18,7 @@ class Command(BaseCommand):
 
     #: normal verbosity level
     v_normal = 1
+    #: verbosity level for the current run; defaults to 1 / normal
     verbosity = v_normal
 
     def add_arguments(self, parser):
@@ -46,7 +45,7 @@ class Command(BaseCommand):
         if source_ids:
             hathi_vols = hathi_vols.filter(source_id__in=source_ids)
 
-        stats = {"updated":0, "unchanged":0, "missing_data":0}
+        stats = {"updated": 0, "unchanged": 0, "missing_data": 0}
 
         for digwork in hathi_vols:
             try:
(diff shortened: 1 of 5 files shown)


ppa/archive/management/commands/update_hathi_pagecounts.py
```diff
@@ -0,0 +1,87 @@
...
+
+from ppa.archive.models import DigitizedWork
+
+
+class Command(BaseCommand):
+    """Update database page counts for non-excerpted HathiTrust digitized items.
+    By default, runs on all non-excerpted, public HathiTrust items.
+    """
+
+    help = __doc__
```
laurejt, line 19: Is the class attribute `help` used?

ppa/archive/management/commands/update_hathi_pagecounts.py
```diff
@@ -0,0 +1,87 @@
...
+    v_normal = 1
+    verbosity = v_normal
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "source_ids", nargs="*", help="List of specific items to update (optional)"
+        )
+
+    def handle(self, *args, **kwargs):
+        verbosity = kwargs.get("verbosity", self.v_normal)
```
laurejt, line 31:
```suggestion
        self.verbosity = kwargs.get("verbosity", self.verbosity)
```

ppa/archive/management/commands/update_hathi_pagecounts.py
```diff
@@ -0,0 +1,87 @@
...
+                        change_message=f"Recalculated page count (was {old_page_count}, "
+                        + f"now {digwork.page_count})",
+                        action_flag=CHANGE,
+                    )
+
+                else:
+                    stats["unchanged"] += 1
+
+            except storage_exceptions.ObjectNotFoundException:
+                if verbosity >= self.v_normal:
```
laurejt, line 75:
```suggestion
                if self.verbosity >= self.v_normal:
```

ppa/archive/management/commands/update_hathi_pagecounts.py
```diff
@@ -0,0 +1,87 @@
...
+                    self.stderr.write(
+                        self.style.WARNING(f"Pairtree data for {digwork} not found")
+                    )
+                stats["missing_data"] += 1
+
+        # report a summary of what was done
+        if verbosity >= self.v_normal:
```
laurejt, line 82:
```suggestion
        if self.verbosity >= self.v_normal:
```

ppa/archive/management/commands/update_hathi_pagecounts.py
```diff
@@ -0,0 +1,87 @@
...
+        hathi_vols = DigitizedWork.objects.filter(
+            source=DigitizedWork.HATHI,
+            item_type=DigitizedWork.FULL,
+            status=DigitizedWork.PUBLIC,
+        )
+        # if source ids are specified, limit to those records only
+        if source_ids:
+            hathi_vols = hathi_vols.filter(source_id__in=source_ids)
+
+        stats = defaultdict(int)
```
laurejt, line 49: For clarity, might it be worth fully declaring the desired dict structure upfront?
```suggestion
        stats = {"update":0, "unchanged":0}
```
This would also remove the use of `defaultdict`

ppa/archive/management/commands/update_hathi_pagecounts.py
```diff
@@ -0,0 +1,87 @@
...
+                    )
+                stats["missing_data"] += 1
+
+        # report a summary of what was done
+        if verbosity >= self.v_normal:
+            self.stdout.write(
+                f"""Volumes with updated page count: {stats['updated']:,}
+    Page count unchanged: {stats['unchanged']:,}
+    Missing pairtree data: {stats['missing_data']:,}"""
+            )
```
laurejt, line 87: I'm unfamiliar with how spacing for logging is handled, but might it be better to use explicit tabs `\t` instead of spaces? This would improve the file formatting and remove the dependency on triple quotes.

laurejt (changes requested):
See my comments above. Before merging, the verbosity issue should be resolved.

rlskoeser:
@laurejt thank you for the careful review. I've incorporated your suggestions & addressed your comments, and also made the change I was planning, so that the `count_pages` method no longer saves the record.

laurejt (approved):

Issue #598, opened 2024-03-04
update HathiTrust rsync


mnaydan:

rlskoeser:
No longer needed

Issue #428, opened 2021-07-29
As a developer, I want a script to update all HathiTrust content so that I can refresh locally cached data with OCR improvements and other changes.


rlskoeser:
As @mnaydan and I discovered in testing excerpts, when we found text and page image mismatches, (and as I've suspected for a while), we should probably be doing regular rsync updates of our HathiTrust pairtree data to get any changes and (presumably) deletions.

the manual version looks something like this:
- get list of all hathi source ids in the database
- create a text file with one id per line
- use the one off script to generate a path list for rsync (per [hathitrust instructions](https://github.com/hathitrust/datasets/wiki/Dataset-rsync-instructions))
- run rsync with the path file list generated, using the options hathitrust specifies: `rsync --copy-links --recursive --times --verbose  --itemize-changes --files-from=/tmp/rsync_path_list.txt datasets.hathitrust.org::ht_text_pd .`. (itemize changes was a local addition so we could see specifically what files changed; I didn't use the delete option but we will need to add it). Note: might need to do something about permissions or file ownership. When I ran this in production, the permissions on some files did not allow the deploy user to set the time attributes (likely because files were created via the admin interface, so they were owned by apache instead of deploy; should also check that the rsync command generated by admin uses the same options).
- After rsync completes, reindex pages to get updated content: `python manage.py index_pages`

We'll need to do these steps in production because we know we have content out of date, and it affects some of the excerpts.

We probably want to automate this at some point — I think if we do that, it would help out if there are any deletions in our dataset, but I'm not clear on all the steps we'd need.

Currently thinking it might look like this:
- custom manage command that can be run as a cron job 
   - gets a unique list of hathitrust source ids from the database
   - calls rsync with those ids, similar to the way we call rsync for adding new records now
   - identifies any records where the content was updated 
       - reindexes pages for updated items
   - identifies any records that have been deleted (not sure how! file contents will be gone, but hopefully there's some way to get rsync to tell us what it removed)
      - mark database record as suppressed; add something to the note field and/or a log entry to document why/when it was suppressed (suppressing will take care of removing it from the index)
     - alert ppa admins (via email? slack?) that an item has been removed — they should be aware; they may need to find and import a different copy of that volume

[...]

PR #614, opened 2024-03-07
Improve hathi_rsync manage command so it can sync and report on all hathi content

rlskoeser:
relates most directly to #428 but also relevant to most of the rsync items aggregated by #598

changes in this PR:
- improves the existing `hathi_rsync` command so it can be run to sync content for all non-suppressed hathi records in PPA
  - fix importer rsync logging so it does not output all source ids when syncing more than a handful of records
  - add an optional output dir for rsync itemized output
- parse rsync itemized output to identify paths that were updated due to a change; convert path to htid and generate a report of all updated htids

I have tested this locally using an rsync from staging server to my local development server and it seems to be working for the scenarios I've tried.

I haven't worked on unit tests related to this change yet, in part because I think this may not be the final form we want - in part because I think this will be one piece in a larger pipeline of updates triggered by content being updated through rsync.

My rsync output parsing does not do any checking for deleted files; that's a different scenario than we're currently concerned about, but we might want to think about that at some point. (... and looks like I made some notes on that on #428)

### DIFF
--- a/ppa/archive/import_util.py
+++ b/ppa/archive/import_util.py
@@ -159,7 +159,14 @@ class DigitizedWorkImporter:
 
 class HathiImporter(DigitizedWorkImporter):
     """Logic for creating new :class:`~ppa.archive.models.DigitizedWork`
-    records from HathiTrust. For use in views and manage commands."""
+    records from HathiTrust. For use in views and manage commands.
+
+    :param list source_ids: list of HathiTrust source ids (htid) to
+        synchronize (optional)
+    :param bool rsync_output: determines whether rsync itemized report
+        is enabled (default: False)
+    :param str output_dir: base directory for rsync output file (optional)
+    """
 
     #: rsync error
     RSYNC_ERROR = 4
@@ -178,10 +185,11 @@ class HathiImporter(DigitizedWorkImporter):
         }
     )
 
-    def __init__(self, source_ids=None, rsync_output=False):
+    def __init__(self, source_ids=None, rsync_output=False, output_dir=""):
         super().__init__(source_ids)
         # track whether (and how much) rsync output is desired
         self.rsync_output = rsync_output
+        self.output_dir = output_dir
 
     def filter_invalid_ids(self):
         """Remove any ids that don't look valid. At minimum, must
@@ -247,7 +255,15 @@ class HathiImporter(DigitizedWorkImporter):
     def rsync_data(self):
         """Use rsync to retrieve data for the volumes to be imported."""
 
-        logger.info("rsyncing pairtree data for %s", ", ".join(self.source_ids))
+        # limit the number of ids included in the log message
+        log_detail = ""
+        rsync_count = len(self.source_ids)
+        if rsync_count <= 12:
+            log_detail = ", ".join(self.source_ids)
+        else:
+            log_detail = "%d volumes" % rsync_count
+
+        logger.info("rsyncing pairtree data for %s", log_detail)
 
         # create temp file with list of paths to synchronize
         with tempfile.NamedTemporaryFile(
@@ -271,8 +287,9 @@ class HathiImporter(DigitizedWorkImporter):
             # if rsync output requested, include itemize and log fileargs
             output_opts = ""
             if self.rsync_output:
-                outputfilename = "ppa_hathi_rsync_%s.log" % datetime.now().strftime(
-                    "%Y%m%d-%H%M%S"
+                outputfilename = os.path.join(
+                    self.output_dir,
+                    "ppa_hathi_rsync_%s.log" % datetime.now().strftime("%Y%m%d-%H%M%S"),
                 )
                 # output requested: always log content to a file
                 output_opts = "--log-file=%s" % outputfilename
@@ -292,7 +309,7 @@ class HathiImporter(DigitizedWorkImporter):
                 subprocess.run(args=rsync_cmd.split(), check=True)
             except subprocess.CalledProcessError as err:
                 logger.error(
-                    "HathiTrust rsync failed — %s / command: %s"
+                    "HathiTrust rsync failed — %s / command: %s"
                     % (self.RSYNC_RETURN_CODES[err.returncode], rsync_cmd)
                 )
 
--- a/ppa/archive/management/commands/hathi_rsync.py
+++ b/ppa/archive/management/commands/hathi_rsync.py
@@ -1,4 +1,8 @@
-from django.core.management.base import BaseCommand, CommandError
+import os.path
+from datetime import datetime
+
+from django.core.management.base import BaseCommand
+from pairtree import path2id
 
 from ppa.archive.import_util import HathiImporter
 from ppa.archive.models import DigitizedWork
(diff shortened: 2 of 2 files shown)


ppa/archive/import_util.py
```diff
@@ -247,7 +255,15 @@ def pairtree_paths(self):
...
-        logger.info("rsyncing pairtree data for %s", ", ".join(self.source_ids))
+        # limit the number of ids included in the log message
+        log_detail = ""
+        rsync_count = len(self.source_ids)
+        if rsync_count <= 12:
+            log_detail = ", ".join(self.source_ids)
+        else:
+            log_detail = "%d volumes" % rsync_count
+
+        logger.info("rsyncing pairtree data for %s", log_detail)
```
rlskoeser, line 266: previously this was logging _all_ of the source ids, which worked for small-batches of imports or rsync but is obviously unhelpful when it's all the hathitrust records in ppa

laurejt: What's the reasoning for setting the default to 12 specifically?

ppa/archive/import_util.py
```diff
@@ -292,7 +309,7 @@ def rsync_data(self):
                 subprocess.run(args=rsync_cmd.split(), check=True)
             except subprocess.CalledProcessError as err:
                 logger.error(
-                    "HathiTrust rsync failed — %s / command: %s"
+                    "HathiTrust rsync failed — %s / command: %s"
```
laurejt, line 312: Did this line actually change? Or, is it just the diff failing.

ppa/archive/import_util.py
rlskoeser: @laurejt totally arbitrary! Just a guess at what might be reasonable. I don't think we want more than a few lines of logging worth of ids, it's not particularly meaningful and it's too noisy. (I did not actually check how this looks in the log files). Could change to  10 or 15 if you think that would be a less distracting number. 😁

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -26,22 +30,64 @@ def handle(self, *args, **kwargs):
...
+        )
 
-        # NOTE: if htid is specified, should we verify that it's
-        # in the db and not suppressed? (should import first if not)
+        # if htids are specified via parameter, use them to filter
+        # the queryset, to ensure we only sync records that are
+        # in the database and not suppressed
+        if htids:
+            digworks = digworks.filter(source_id__in=htids)
+        # NOTE: report here on any skipped ids?
```
laurejt, line 43: It seems worth logging any skipped ids (either because they don't exist or are suppressed)

ppa/archive/import_util.py
```diff
@@ -247,7 +255,15 @@ def pairtree_paths(self):
     def rsync_data(self):
         """Use rsync to retrieve data for the volumes to be imported."""
 
-        logger.info("rsyncing pairtree data for %s", ", ".join(self.source_ids))
+        # limit the number of ids included in the log message
+        log_detail = ""
+        rsync_count = len(self.source_ids)
+        if rsync_count <= 12:
```
laurejt, line 261: Suggestion: switch to 10 since it's typically interpreted as less specific than 12.
```suggestion
        if rsync_count <= 10:
```

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -26,22 +30,64 @@ def handle(self, *args, **kwargs):
...
 
-        self.stdout.write(
-            self.style.SUCCESS("Synchronizing data for %d records" % len(htids))
-        )
-        # even if verbosity is zero we want an output file
+        # generate a list of unique source ids from the queryset
+        hathi_ids = digworks.values_list("source_id", flat=True).distinct()
```
laurejt, line 46: Consider renaming the variable `hathi_ids`, it's easy to confuse this with `htids`
```suggestion
        working_htids = digworks.values_list("source_id", flat=True).distinct()
```

ppa/archive/import_util.py
```diff
@@ -178,10 +185,11 @@ class HathiImporter(DigitizedWorkImporter):
         }
     )
 
-    def __init__(self, source_ids=None, rsync_output=False):
+    def __init__(self, source_ids=None, rsync_output=False, output_dir=""):
         super().__init__(source_ids)
         # track whether (and how much) rsync output is desired
         self.rsync_output = rsync_output
+        self.output_dir = output_dir
```
laurejt, line 192: For the purposes of initialization, it seems like there should be an additional constraint that if `rsync_output` is set then `output_dir` must also be specified (and exist).

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -26,22 +30,64 @@ def handle(self, *args, **kwargs):
...
+        self.stdout.write("Synchronizing data for %d records" % len(hathi_ids))
+        # we always want itemized rsync output, so we can report
+        # on which volumes were updated
         htimporter = HathiImporter(
-            source_ids=htids, rsync_output=self.verbosity or True
+            source_ids=hathi_ids, rsync_output=True, output_dir="/tmp"
```
laurejt, line 51: Is `/tmp` a good output directory for this?

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -26,22 +30,64 @@ def handle(self, *args, **kwargs):
...
         )
         logfile = htimporter.rsync_data()
-        self.stdout.write(self.style.SUCCESS("rsync output is in %s" % logfile))
+
+        # read the rsync itemized output to identify records where file
+        # sizes changed
```
laurejt, line 56: Is it possible for a file to change in a way that matters to use without its size reflecting this?

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -26,22 +30,64 @@ def handle(self, *args, **kwargs):
...
+                # to a change in size, use the path to determine the hathi id
+                if " >f.s" in line:
+                    # rsync itemized output is white-space delimited;
+                    # last element is the filename that was updated
+                    filename = line.rsplit()[-1].strip()
+                    # we only care about zip files and mets.xml files
+                    if not filename.endswith(".zip") and not filename.endswith(".xml"):
+                        continue
+                    # reconstruct the hathi id from the filepath
+                    ht_prefix, pairtree_dir = filename.split("/pairtree_root/")
```
laurejt, line 70:
```suggestion
                    ht_prefix, pairtree_dir = filename.split("/pairtree_root/", 1)
```

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -26,22 +30,64 @@ def handle(self, *args, **kwargs):
...
+                    # get the directory one level up from the updated file
+                    pairtree_id = os.path.dirname(os.path.dirname(pairtree_dir))
+                    # use pairtree to determine the id based on the path
+                    # (handles special characters like those used in ARKs)
+                    htid = f"{ht_prefix}.{path2id(pairtree_id)}"
+                    updated_ids.add(htid)
+
+        # should this behavior only be when updating all?
+        # if specific htids are specified on the command line, maybe report on them only?
```
laurejt, line 79: It seems like it'd always be useful to know *which* files were updated by rsync, since it's not guaranteed that any are.

laurejt (commented):
This looks good overall, but I have one concern about whether a change in size encompasses all changes of interest.

ppa/archive/import_util.py
rlskoeser: there was a weird character here! I removed it but you can't tell what the change is, I guess it was a different kind of whitespace

ppa/archive/import_util.py
rlskoeser: previously it was putting the file in the current working directory - but you're right, better to be explicit

ppa/archive/management/commands/hathi_rsync.py
rlskoeser: Oh, I meant to look into that and forgot. If and when we set this functionality to run as a cron job, that's not a good place. In staging/production, maybe it could go into the logging directory. It might be good enough for now, while we're assessing.

rlskoeser:
@laurejt thanks for the great feedback. Appreciate the suggestion for the variable name `working_htids`, much clearer!

updated as follows:
- report on skipped htids if specific ids are passed as a parameter
- now using a `tempfile.TemporaryDirectory` instead of `/tmp`
- generating a csv report of updated files that preserves a lot more of the detail from rsync; may not be what we want ultimately, but could be helpful as we're investigating how rsync helps us with the other problems.
- added and tested validation for the output dir parameter for the rsync logging

ppa/archive/import_util.py
```diff
@@ -185,10 +188,19 @@ class HathiImporter(DigitizedWorkImporter):
...
         super().__init__(source_ids)
         # track whether (and how much) rsync output is desired
         self.rsync_output = rsync_output
+        # if rsync output is enabled, output directory is required
+        if self.rsync_output:
+            if output_dir is None:
+                raise ValueError("output_dir is required when rsync_output is enabled")
+            elif not os.path.isdir(output_dir):
+                raise ValueError(
+                    "rsync output dir %s is not an existing directory", output_dir
```
laurejt, line 201:
```suggestion
                    f"rsync output dir {output_dir} is not an existing directory"
```

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -40,52 +42,88 @@ def handle(self, *args, **kwargs):
...
+                    self.style.NOTICE(
+                        "Some ids not found in public HathiTrust volumes; skipping %s"
+                        % " ".join(skipped_htids)
+                    )
+                )
+
+        # bail out if there's nothing to do
+        # (e.g., explicit htids only and none valid)
+        if not working_htids:
+            return
```
laurejt, line 63: It seems worth logging/reporting that there's nothing to synchronize before exiting

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -40,52 +42,88 @@ def handle(self, *args, **kwargs):
...
+        # if htids were explicitly specified, report if any are skipped
+        if htids:
+            skipped_htids = set(htids) - set(working_htids)
+            if skipped_htids:
+                self.stdout.write(
+                    self.style.NOTICE(
+                        "Some ids not found in public HathiTrust volumes; skipping %s"
+                        % " ".join(skipped_htids)
+                    )
+                )
```
laurejt, line 58: Perhaps report the total # of volumes skipped? See also comment for line 63.

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -40,52 +42,88 @@ def handle(self, *args, **kwargs):
...
-                if " >f.s" in line:
-                    # rsync itemized output is white-space delimited;
+                # check for a line indicating that a file was updated
+                if " >f" in line:
+                    # rsync itemized output is white-space delimited
+                    parts = line.split()
                     # last element is the filename that was updated
-                    filename = line.rsplit()[-1].strip()
+                    filename = parts[-1]
+                    # itemized info flags preced the filename
```
laurejt, line 86:
```suggestion
                    # itemized info flags precede the filename
```

ppa/archive/management/commands/hathi_rsync.py
```diff
@@ -40,52 +42,88 @@ def handle(self, *args, **kwargs):
...
+                        }
+                    )
 
         # should this behavior only be when updating all?
         # if specific htids are specified on the command line, maybe report on them only?
-        if updated_ids:
-            outfilename = "ppa_rsync_updated_htids_%s.txt" % datetime.now().strftime(
+        if updated_files:
+            outfilename = "ppa_rsync_changes_%s.csv" % datetime.now().strftime(
                 "%Y%m%d-%H%M%S"
```
laurejt, line 115: To improve readability
```suggestion
            outfilename = "ppa_rsync_changes_{time}.csv".format(
                time=datetime.now().strftime("%Y%m%d-%H%M%S")
```

ppa/archive/tests/test_import_util.py
```diff
@@ -269,6 +268,26 @@ def test_rsync_data(self, mocksubprocess):
...
+    assert htimporter.output_dir is None
+
+    # rsync output requested with no output dir
+    with pytest.raises(ValueError, match="output_dir is required"):
+        HathiImporter(rsync_output=True)
+
+    # rsync output requested with non-existent output dir
+    with pytest.raises(ValueError, match="not an existing directory"):
+        # rsync output requested with invalid output dir
+        HathiImporter(rsync_output=True, output_dir="/tmp/foo/bar")
```
laurejt, line 284: So CodeFactor won't complain (there's no reason `tmp` needs to be mentioned)
```suggestion
        HathiImporter(rsync_output=True, output_dir="/foo/bar/baz")
```

laurejt (approved):

ppa/archive/management/commands/hathi_rsync.py
rlskoeser: Ah, good catch. In earlier versions it output "synchronizing data for 0 records" but I've changed the order.

PR #624, opened 2024-03-20
Feature/collect version labels

laurejt:
This contains scripts and accompanying documentation/data for collecting version information from the HathiTrust catalog public-facing volume pages.


scripts/README.md
```diff
@@ -0,0 +1,16 @@
+# PPA Scripts
+
+This directory contains stand-alone scripts associated with the Princeton
+Prosody Archive that are not part of the web application proper.
+
+At this time, these scripts do not have any additional requirements.
+
+## HathiTrust "Version" Timestamps
+This script extracts and saves the version timestamp information for each
+excerpt record in PPA.
```
rlskoeser, line 10: Is this accurate?

```suggestion
This script extracts and saves the version timestamp information from the
public HathiTrust interface for PPA excerpt records (based on an exported report).
```

scripts/get_version_labels.py
```diff
@@ -0,0 +1,76 @@
...
+        version_label = re.findall(re_pattern, r.text)
+        if version_label:
+            return version_label[0]
+        else:
+            print(f"Warning: {htid} missing versionLabel!")
+    else:
+        print(f"Warning: bad/unexpected response")
+
+
+def get_version_labels(htids, wait_time=3):
```
rlskoeser, line 31: fwiw, there's a python `ratelimit` library I've used before that can apply rate limiting by adding a decorator to a method. (not needed here, just sharing in case it's of interest)

How did you determine the 3 second delay? Is this a HT rate limit?

scripts/get_version_labels.py
```diff
@@ -0,0 +1,76 @@
...
+import datetime
+import requests
+
+
+def get_version_label(htid):
+    """
+    Extract the HathiTrust "version label" from a record's catalog page.
+    Returns the corresponding timestamp, returns None if the HTTP request fails.
+    """
+    re_pattern = r'HT.params.versionLabel = "([^"]+)";'
```
rlskoeser, line 17: I think this is worth a brief comment to explain what we're looking for. Here's my attempt, feel free to edit.
```suggestion
    # a script block in HT record page includes a number of HT.params including version timestamp
    re_pattern = r'HT.params.versionLabel = "([^"]+)";'
```

scripts/get_version_labels.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Get htids
+    htids = []
+    with open(in_tsv) as reader:
+        reader.readline() # skip header
+        for line in reader:
+            fields = line.rstrip().split('\t')
+            if len(fields) < 15:
+                continue
+            htid = fields[15]
```
rlskoeser, line 68: I usually use the python `csv` module for something like this - I presume it would works fine with tsv if you set the delimiter to tab. That way you can use a dict reader and access the htid field based on the label rather than an index.

rlskoeser (approved):
Thanks for working on this, everything looks good to me.  A couple of FYIs and minor suggestions to improve documentation a bit (edit/reject as you see fit).

Merge it whenever you're ready!

scripts/get_version_labels.py
laurejt: It was an arbitrary default. That said, there is a [robots.txt](https://babel.hathitrust.org/robots.txt) that specifies a 1 second time delay, so I'll change the default to that.

scripts/get_version_labels.py
laurejt: I think I'm going to modify the script to just take a file as input / create a modified version for our current purposes. Then we can just assume a simple .txt file with one htid per line (and no header). This lets me clean up this code so it's not having to handle the specific idiosyncrasies of this google sheet.

rlskoeser: that sounds like a good simplification, and easy to generate that list from a variety of sources

rlskoeser (approved):
Looks great!

PR #653, opened 2024-05-15
Add developer notes for updating hathitrust, reindexing, and exporting text corpus

rlskoeser:
notes from our working session; please suggest/make edits if there's anything you want to clarify

### DIFF
--- a/DEVELOPERNOTES.rst
+++ b/DEVELOPERNOTES.rst
@@ -39,6 +39,38 @@ Solr changes not reflected in search results? ``solrconfig.xml`` must be
 updated in Solr's main directory: ``solr/server/solr/[CORE]/conf/solrconfig.xml``
 
 
+Updating HathiTrust records and generating a fresh text corpus
+--------------------------------------------------------------
+
+These command should be run on the production server as the deploy user
+with the python virtual environment activated.
+
+Update all HathiTrust documents with rsync::
+
+    python manage.py hathi_rsync
+
+This file will generate a csv report of the files that were updated.
+Use the resulting file to get a list of ids that need to be indexed:
+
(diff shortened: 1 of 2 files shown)


DEVELOPERNOTES.rst
```diff
@@ -39,6 +39,38 @@ Solr changes not reflected in search results? ``solrconfig.xml`` must be
 updated in Solr's main directory: ``solr/server/solr/[CORE]/conf/solrconfig.xml``
 
 
+Updating HathiTrust records and generating a fresh text corpus
+--------------------------------------------------------------
+
+These command should be run on the production server as the deploy user
```
laurejt, line 45:
```suggestion
These commands should be run on the production server as the deploy user
```

DEVELOPERNOTES.rst
```diff
@@ -39,6 +39,38 @@ Solr changes not reflected in search results? ``solrconfig.xml`` must be
...
+with the python virtual environment activated.
+
+Update all HathiTrust documents with rsync::
+
+    python manage.py hathi_rsync
+
+This file will generate a csv report of the files that were updated.
+Use the resulting file to get a list of ids that need to be indexed:
+
+    cut -f 1 -d, ppa_rsync_changes_[TIMESTAMP].csv | sort | uniq | tail -n +2 > htids.txt
```
laurejt, line 55: Just want to verify: there are no HathiTrust ids with commas. If there are any IDs with commas, `cut` will split too early for those IDs / rows.

rlskoeser: I don't think we've ever run across any HT ids with commas; the ARKs are the oddest format we get, but at least they don't have commas.

laurejt (approved):

PR #654, opened 2024-06-04
Page indexing refactor

rlskoeser:
This refactors the code for page indexing to setup work for the EEBO import.

Previously there were separate methods for Gale and HathiTrust page index data, with substantial overlap in common logic. This refactor implements source-specific page data generators, which are then consumed and updated with the common logic (like excerpt handling) and fields needed for all page records.

Also adds an efficiency improvement: when indexing pages for an excerpt, we don't need to keep iterating once we've got the data for the range we care about.

I also updated one test that has caused problems before, where we had some hard-coded solr page index data. Now it takes advantage of the common page field logic, so if page indexing logic or fields change it will automatically get those updates.

### DIFF
--- a/ppa/archive/models.py
+++ b/ppa/archive/models.py
@@ -1228,6 +1228,10 @@ class Page(Indexable):
 
         # get page span from digitized work, to handle excerpts
         page_span = digwork.page_span
+        # if indexing an excerpt, determine highest page to be indexed
+        max_page = None
+        if page_span:
+            max_page = max(page_span)
         # index id is used to group work and pages; also fallback for cluster id
         # for works that are not part of a cluster
         digwork_index_id = digwork.index_id()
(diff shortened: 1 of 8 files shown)


ppa/archive/models.py
```diff
@@ -1206,126 +1206,69 @@ def index_item_type(cls):
...
     @classmethod
-    def page_index_data(cls, digwork):
+    def page_index_data(cls, digwork, gale_record=None):
         """Get page content for the specified digitized work from Hathi
-        pairtree and return data to be indexed in solr."""
+        pairtree and return data to be indexed in solr.
 
-        # TODO: how to share common fields/logic across sources?
+        Takes an optional Gale item record as returned by the Gale API,
+        to avoid loading API content twice. (Used on import)"""
```
laurejt, line 1214: Method description should be updated to reflect that this method is used for both HathiTrust and Gale items (and in the future EEBO).

PR #671, opened 2024-09-19
Feature/gale local ocr

laurejt:


ppa/archive/gale.py
```diff
@@ -12,6 +12,22 @@
 logger = logging.getLogger(__name__)
 
 
+def get_local_ocr(item_id, page_num):
+    """
+    Get local OCR page text for specified page of a single Gale volume
+    """
```
rlskoeser, line 18: might be helpful to add a bit more context/info here in comments, here's a suggestion:
```suggestion
    """
    # Pull local OCR text for Gale content if available. Requires a base directory to be configured and assumes that
    # files are organized following the convention of stub directories based on every third number from the
    # Gale ID and then named based on item id and page number, matching the image file names provided by Gale.
```

ppa/archive/gale.py
```diff
@@ -192,14 +208,19 @@ def get_item_pages(self, item_id, gale_record=None):
         # iterate through the pages in the response
         for page in gale_record["pageResponse"]["pages"]:
             page_number = page["pageNumber"]
-            # page label (original page number) should be set in folioNumber,
-            # but is not set for all volumes; fallback to page number
-            # converted to integer to drop leading zeroes
-            page_label = page.get("folioNumber", int(page_number))
+            tags = []
```
rlskoeser, line 211: suggest adding a comment here with a bit more details

```suggestion
            tags = []
            # get higher quality local OCR text content if possible, with fallback to Gale OCR
            # set a tag so we have some way to determine when local OCR is present
            
```

ppa/archive/gale.py
```diff
@@ -192,14 +208,19 @@ def get_item_pages(self, item_id, gale_record=None):
...
+            except FileNotFoundError as e:
+                ocr_text = page.get("ocrText")  # some pages have no text
+                logger.warning(f'Local OCR not found for {item_id} {page_number}')
+
             info = {
                 "page_id": page_number,
-                "content": page.get("ocrText"),  # some pages have no text
-                "label": page_label,
+                "content": ocr_text,
+                "label": page.get("folioNumber"),
```
rlskoeser, line 222: This one also seems worth commenting (thanks again for catching this!). Maybe something like this?

```suggestion
                # don't set label when we don't have one, so we get fall-back behavior in common page indexing code
                "label": page.get("folioNumber"),
```

ppa/archive/gale.py
```diff
@@ -192,14 +208,19 @@ def get_item_pages(self, item_id, gale_record=None):
         # iterate through the pages in the response
         for page in gale_record["pageResponse"]["pages"]:
             page_number = page["pageNumber"]
```
rlskoeser, line 210: could just use this value directly instead of assigning to a local variable, now that we're only using it one place

rlskoeser (approved):
Looks great! The tests are nice, and thanks for catching that inferred page label issue.

I'm suggesting adding a few more comments and I think you should do something about the path issue that the AI code review flagged, but I think that's all that's needed.

ppa/archive/gale.py
```diff
@@ -12,6 +12,22 @@
...
+    """
+    ocr_dir = getattr(settings, "GALE_LOCAL_OCR", None)
+    if not ocr_dir:
+        raise ImproperlyConfigured(
+            "GALE_LOCAL_OCR configuration is required for indexing Gale page content"
+        )
+
+    stub_dir = item_id[::3][1:]  # Following conventions set in ppa-nlp
+    ocr_txt_fp = f"{ocr_dir}/{stub_dir}/{item_id}/{item_id}_{page_num}0.txt"
+    with open(ocr_txt_fp) as reader:
```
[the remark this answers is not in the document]
laurejt: Yeah, using `os.path.join` seems like. abetter idea to me.

rlskoeser (approved):
Thanks for adding so much detail on the local ocr setup, really helpful to have here.

The other revisions look good.

Issue #665, opened 2024-08-16
write code so that all note types in eebo-xml appear at bottom of page instead of inline


mnaydan:

rlskoeser:
@jerielizabeth I was able to get all the EEBO-TCP pages indexed with the new logic, but I had to do it a few ids at a time. When I try to index all of them, either by id or by source, the indexing hangs at some point. This didn't happen when I indexed locally (although I thought I saw one error which I then couldn't duplicate); I'll see if I can isolate whether a particular volume is causing the problem. Maybe Laure will spot the problem in her code review.

This issue has some comments from when Mary and I were investigating the EEBO-TCP xml:
https://github.com/Princeton-CDH/ppa-django/issues/659

It includes some examples of texts with notes of various kinds with links so you can see how they display on the U Michigan site and the text content around and in the note.

Here are a couple of searches on the test site for content near footnotes where you can now see a footnote indicator rather than the inline footnote contents (as was happening previously):
https://test-prosody.cdh.princeton.edu/archive/A66045/?query=%22handling+of+that+subject%22
https://test-prosody.cdh.princeton.edu/archive/CW0116702026/?query=%22Exact+propriety+of+word+and+thought+%22

jerielizabeth:
I tested the functionality using a jsonl file provided by @rlskoeser for item [A68619](http://name.umdl.umich.edu/A68619.0001.001) in EEBO (PPA link: https://test-prosody.cdh.princeton.edu/archive/A68619/) 

Examples of working as expected include the following pages:

- p. 137: [PPA](https://test-prosody.cdh.princeton.edu/archive/A68619/?query=%22by+vs+the+Ringleader%22) | [UMich](https://quod.lib.umich.edu/e/eebo2/A68619.0001.001?rgn=main;view=fulltext#fnDLPS3) | 4 expected citations in the text, text for which does appear at the bottom of the "text" field in the jsonl file.
- p 158: [PPA](https://test-prosody.cdh.princeton.edu/archive/A68619/?query=%22Or+when+we+speake+by+manner+of+pleasantery%22) | [UMich](https://quod.lib.umich.edu/e/eebo2/A68619.0001.001?rgn=main;view=fulltext#fnDLPS40) | example of footnote content being outside the displayed text as expected.
[...]

PR #684, opened 2024-10-22
Revise eebo-tcp page text logic to put notes after main page text

rlskoeser:
ref #665

Revise eebo-tcp page text generation to identify and collect notes on the text, display note markers inline, and note contents after main page content

### DIFF
--- a/ppa/archive/eebo_tcp.py
+++ b/ppa/archive/eebo_tcp.py
@@ -28,8 +28,60 @@ class Page(xmlmap.XmlObject):
     # use following axis to find all text nodes following this page beginning
     text_contents = xmlmap.StringListField("following::text()")
 
+    # notes on or after this page, for managing footnote text
+    notes = xmlmap.NodeListField("following::NOTE", xmlmap.XmlObject)
+
     def __repr__(self):
         return f"<Page {self.number or '-'} ({self.section_type})>"
+
+    note_markers = [
+        "*",
+        "†",
+        "‡",
+        "§",
+        "**",
+        "††",
+        "‡‡",
+        "§§",
+        "***",
+        "†††",
+        "‡‡‡",
+        "§§§",
+    ]
+
+    def text_inside_note(self, text):
+        """check if a text element occurs within a NOTE element; if so,
+        return the note element"""
+
+        # bail out if there are no notes following this page
+        if not self.notes:
+            return None
+
+        within_note = None
+        # check if this text is directly inside a note tag
+        parent = text.getparent()
+        if parent.tag == "NOTE" and not text.is_tail:
+            within_note = parent
+        # otherwise, check if text is nested under a note tag
+        else:
+            for ancestor in parent.iterancestors():
+                if ancestor.tag == "NOTE":
+                    within_note = ancestor
+                    break
+
+        return within_note
+
+    def note_index(self, note):
+        # given a note element, determine the 0-based index on this page
+
+        # use a for loop so we can bail out once we get a match
+        for i, n in enumerate(self.notes):
+            if n.node == note:
+                return i
+
+        # in normal use the note should be found; raise an exception
+        # if it is not so this will fail loudly
+        raise ValueError
 
     def page_contents(self):
         """generator of text strings between this page beginning tag and
@@ -38,16 +90,40 @@ class Page(xmlmap.XmlObject):
         # strictly speaking we are returning lxml "smart strings"
         # (lxml.etree._ElementUnicodeResult)
 
+        # collect any notes and include after main page text contents
+        notes = []
+
         # iterate and yield text following the current page
         # break until we hit the next page beginning
         for i, text in enumerate(self.text_contents):
             parent = text.getparent()
 
+            # determine if this text falls inside a note tag
+            within_note = self.text_inside_note(text)
+            if within_note is not None:
+                # and if so, which one
+                note_index = self.note_index(within_note)
+                # if this is the first text for this note,
+                # add a marker inline with the text AND the note
+                # index equals length, start a new note at the end of the list of notes
+                if len(notes) == note_index:
+                    # some note tags have an N attribute; use if present
+                    # otherwise, use marker from our list based on sequence
+                    note_marker = within_note.get("N", self.note_markers[note_index])
+                    yield note_marker
+                    notes.append(f"\n{note_marker} ")
+
+                # add text to the appropriate note
+                notes[note_index] = f"{notes[note_index]}{text}"
+
+                # skip to next loop
+                continue
+
             # lxml handles text between elements as "tail" text;
             # the parent of the tail text is the preceding element
             if text.is_tail and parent.tag == "GAP":
                 # if text precedes a GAP tag, include the display content
-                # from the DISP attribute (for now)
+                # from the DISP ` (for now)
                 yield text.getparent().get("DISP")
 
             if text.is_tail and parent.tag == "PB":
@@ -58,6 +134,12 @@ class Page(xmlmap.XmlObject):
                     break
 
             yield text
+
+        # if this page includes notes, yield notes after main text content
+        if notes:
+            # yield two blank lines to separate main text content from notes
+            yield "\n\n"
+            yield from notes
 
     divider = "∣"
 
--- a/ppa/archive/eebo_tcp.py
+++ b/ppa/archive/eebo_tcp.py
@@ -27,9 +27,8 @@ class Page(xmlmap.XmlObject):
     # page beginning tags delimit content instead of containing it;
     # use following axis to find all text nodes following this page beginning
     text_contents = xmlmap.StringListField("following::text()")
-
-    # notes on or after this page, for managing footnote text
-    notes = xmlmap.NodeListField("following::NOTE", xmlmap.XmlObject)
+    # count following notes as a quick check to bail out of note detection logic
+    has_notes = xmlmap.IntegerField("count(following::NOTE)")
 
     def __repr__(self):
         return f"<Page {self.number or '-'} ({self.section_type})>"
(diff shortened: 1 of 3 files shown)


ppa/archive/tests/test_eebo_tcp.py
```diff
@@ -39,6 +39,72 @@ def test_eebo_tcp_page_contents():
...
+    # text nested under a tag within a note tag (note is ancestor)
+    camden_note_text = text.node.xpath("//NOTE/HI[contains(., 'Cambden')]/text()")[0]
+    assert page.text_inside_note(camden_note_text) is not None
+    # text in a tag that is NOT inside a note tag (note is not parent/ancestor)
+    europe_hi_text = text.node.xpath("//HI[contains(., 'Europe')]/text()")[0]
+    assert page.text_inside_note(europe_hi_text) is None
+
+
+def test_get_note_mark():
+    text = load_xmlobject_from_string(PAGE_WITH_NOTE, eebo_tcp.Text)
```
laurejt, line 71: Will this also pass if we load from an empty or place holder or note-free xml page?

ppa/archive/tests/test_eebo_tcp.py
```diff
@@ -39,6 +39,72 @@ def test_eebo_tcp_page_contents():
...
+    page = text.pages[0]
+    # text directly inside a note tag (note is parent)
+    diatribe_note_text = text.node.xpath("//NOTE[contains(., 'Diatribe')]/text()")[0]
+    assert page.text_inside_note(diatribe_note_text) is not None
+    # text nested under a tag within a note tag (note is ancestor)
+    camden_note_text = text.node.xpath("//NOTE/HI[contains(., 'Cambden')]/text()")[0]
+    assert page.text_inside_note(camden_note_text) is not None
+    # text in a tag that is NOT inside a note tag (note is not parent/ancestor)
+    europe_hi_text = text.node.xpath("//HI[contains(., 'Europe')]/text()")[0]
+    assert page.text_inside_note(europe_hi_text) is None
```
laurejt, line 67: Perhaps a simpler toy example could be used?

ppa/archive/tests/test_eebo_tcp.py
laurejt, line 1: Some page notes testing that appears to be missing:
- Checking that all notes for a page are extracted (i.e. check `len(page.notes)`)
- Checking that specified placeholder markers (i.e. `N` fields) are used when available and that this interacts correctly when there is a mixed type of notes (hopefully this doesn't occur)
- It'd be nice to separate the testing of the collecting of a page's notes from its display, although I don't know how easy that is.

ppa/archive/eebo_tcp.py
```diff
@@ -28,26 +28,100 @@ class Page(xmlmap.XmlObject):
...
+
     def page_contents(self):
         """generator of text strings between this page beginning tag and
         the next one"""
 
         # strictly speaking we are returning lxml "smart strings"
         # (lxml.etree._ElementUnicodeResult)
 
+        # collect any notes and include after main page text contents
+        notes = []
```
laurejt, line 92: is there a reason this doesn't correspond to `self.notes`?

ppa/archive/eebo_tcp.py
```diff
@@ -28,26 +28,100 @@ class Page(xmlmap.XmlObject):
     # use following axis to find all text nodes following this page beginning
     text_contents = xmlmap.StringListField("following::text()")
 
+    # notes on or after this page, for managing footnote text
```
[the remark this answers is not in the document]
rlskoeser: My xmlobject class here is named `Page` but that's somewhat inaccurate, since in tei pages aren't containers but only delimited by `<PB>` tags. This xpath will find _all_ notes after the current page / pb. In a volume with a lot of notes that might be problematic, I'm not sure. I was wondering about looking to see if it's possible to filter the xpath to only find notes after this pb and before the next one, but I didn't go looking.

ppa/archive/eebo_tcp.py
```diff
@@ -28,26 +28,100 @@ class Page(xmlmap.XmlObject):
...
+
+    def get_note_mark(self, i):
+        """Generate a note marker based on note index and :attr:`note_marks`;
+        symbols are used in order and then doubled, tripled, etc as needed."""
+
+        # use modulo to map to the list of available marks
+        mark_index = i % self.num_note_marks
+        # use division to determine how many times to repeat the mark
+        repeat = int(i / self.num_note_marks) + 1
+        return self.note_marks[mark_index] * repeat
```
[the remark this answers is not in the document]
rlskoeser: as we discussed in the demo meeting - closer to the example markers I saw in the xml and I think it's less likely to be confused with other text content

ppa/archive/eebo_tcp.py
```diff
@@ -28,26 +28,100 @@ class Page(xmlmap.XmlObject):
...
+        # bail out if there are no notes following this page
+        if not self.notes:
+            return None
+
+        # check if this text is directly inside a note tag
+        within_note = None
+        # check if this is normal text directly inside a note tag
+        parent = text.getparent()
+        if parent.tag == "NOTE" and text.is_text:
+            within_note = parent
```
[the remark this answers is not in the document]
rlskoeser: I'm not sure what you mean here

ppa/archive/eebo_tcp.py
```diff
@@ -28,26 +28,100 @@ class Page(xmlmap.XmlObject):
...
+        """Generate a note marker based on note index and :attr:`note_marks`;
+        symbols are used in order and then doubled, tripled, etc as needed."""
+
+        # use modulo to map to the list of available marks
+        mark_index = i % self.num_note_marks
+        # use division to determine how many times to repeat the mark
+        repeat = int(i / self.num_note_marks) + 1
+        return self.note_marks[mark_index] * repeat
+
+    def text_inside_note(self, text):
```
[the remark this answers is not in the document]
rlskoeser: Yes, I was initially thinking of it as a boolean (you'll see `within_note` elsewhere). Any suggestions for rename?

ppa/archive/eebo_tcp.py
```diff
@@ -28,26 +28,100 @@ class Page(xmlmap.XmlObject):
...
+            if within_note is not None:
+                # if text is inside a note, determine which one
+                note_index = self.note_index(within_note)
+                # if this is the first text for this note,
+                # add a marker inline with the text AND the note
+                # index equals length, start a new note at the end of the list of notes
+                if len(notes) == note_index:
+                    # some note tags have an N attribute; use if present
+                    # otherwise, use a note mark from our list of symbols
+                    note_mark = within_note.get("N", self.get_note_mark(note_index))
```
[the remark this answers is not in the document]
rlskoeser: Good idea, I'll add a comment

ppa/archive/tests/test_eebo_tcp.py
rlskoeser: I think it would be fine, but you're right that I should test

ppa/archive/eebo_tcp.py
rlskoeser: hmm, I should probably rename those to clarify/differentiate... `self.notes` is the xmlobject reference to all notes on or after the current page; this local `notes` variable is a list of plain text notes contents to be yielded as page content after the main text of the page. Any thoughts on names?

ppa/archive/eebo_tcp.py
```diff
@@ -27,9 +27,8 @@ class Page(xmlmap.XmlObject):
     # page beginning tags delimit content instead of containing it;
     # use following axis to find all text nodes following this page beginning
     text_contents = xmlmap.StringListField("following::text()")
-
-    # notes on or after this page, for managing footnote text
-    notes = xmlmap.NodeListField("following::NOTE", xmlmap.XmlObject)
+    # count following notes as a quick check to bail out of note detection logic
+    has_notes = xmlmap.IntegerField("count(following::NOTE)")
```
laurejt, line 31: Since this is only used for boolean tests, make it a boolean. Otherwise, consider renaming it to reflect that it isn't a boolean.

ppa/archive/eebo_tcp.py
```diff
@@ -64,57 +64,51 @@ def text_inside_note(self, text):
...
-                note_index = self.note_index(within_note)
-                # if this is the first text for this note,
-                # add a marker inline with the text AND the note
-                # index equals length, start a new note at the end of the list of notes
-                if len(notes) == note_index:
-                    # some note tags have an N attribute; use if present
-                    # otherwise, use a note mark from our list of symbols
+                # is this the first text in this note?
+                within_note.xpath(".//text()")
+                # print(f"first note text = {first_text}")
```
laurejt, line 97: Should this be line be deleted or turned into logging?

ppa/archive/eebo_tcp.py
```diff
@@ -64,57 +64,51 @@ def text_inside_note(self, text):
...
+                is_first_text = within_note.xpath(".//text()")[0] == str(text)
+                if is_first_text:
+                    # print(f"hit first note text {text}")
```
laurejt, line 100: Should this be line be deleted or turned into logging?

ppa/archive/tests/test_eebo_tcp.py
```diff
@@ -39,32 +39,53 @@ def test_eebo_tcp_page_contents():
     )
 
 
+# ruff: noqa: E501
```
laurejt, line 42: Given that the line-too-long check has been turned off, why was the `PAGE_WITH_MULTIPLE_NOTES` broken into so many lines?

ppa/archive/tests/test_eebo_tcp.py
```diff
@@ -87,10 +108,10 @@ def test_eebo_tcp_page_contents_notes():
     page_contents = str(text.pages[0])
     # should not display note contents inline
     assert "taken from Sir John Derhan. A headless Carcass," not in page_contents
-    # should display a note marker inline
-    assert "* A headless Carcass" in page_contents
+    # should display a note marker inline using  note marker from the xml
```
laurejt, line 111:
```suggestion
    # should display a note marker inline using note marker from the xml
```

ppa/archive/eebo_tcp.py
rlskoeser: eulxml has a boolean field but it was written for reading and writing boolean values, but I couldn't get it to work here - I think it must assume string values and can't handle boolean or integer values; I'll add a comment

laurejt: Sorry, I keep forgetting this is all eulxml

Issue #673, opened 2024-09-25
Improve Gale indexing speed for local OCR


jerielizabeth:
> Estimating time difference in indexing Gale volumes with current fix:
> 
> 
> 
> Test case: index 10 Gale volumes
> 
> - Volumes: CB0127060085 CB0126086107 CB0127060264 CB0128005380 CB0128002412 CW0115524470 CW0111606033 CW0114828122 CW0125924161 CW0110904426
> 
> - Pages: 1,762 pages
> 
> 
> 
> Timing:
> 
> - Production (original): 11 seconds
> 
> - Test (new): 1 minutes 14 seconds
> 
> 
> 
> The new indexing is *substantially* slower and should be optimized in the future. 

 _Originally posted by @laurejt in [#668](https://github.com/Princeton-CDH/ppa-django/issues/668#issuecomment-2371757347)_

rlskoeser:
I've been doing some testing in staging to see how the revised OCR file structure impacts the time to complete indexing all of our Gale content.

The first time I ran it, it took a little over an hour to complete - a substantial improvement over the ~5 hours I saw most recently in production. I was curious if tigerdata was slowing this down substantially, so I copied the OCR json files to the local filesystem and ran it again - the time was  basically the same as when indexing with OCR files on tigerdata. I suspected there are enough slow downs between filesystem reads and waiting on the Gale API (which can be slow for larger volumes) that cpu / computation isn't the bottleneck, so I tested indexing with 4 processes (even though we only have 2 cpus on this vm), and indexing completed in 30 minutes.

That seems like a reasonable enough improvement to me.  When we set up the cron job we should use `-p 4` to index with 4 processes.

PR #686, opened 2024-10-28
Revise local ocr handling for gale page indexing

rlskoeser:
related to #673 

see this pull request for the script to convert the text files into the json format used https://github.com/Princeton-CDH/ppa-nlp/pull/107

updates since draft version:
- added error handling (adapted from previous implementation)
- updated/adapted documentation 
- updated unit tests

### DIFF
--- a/ppa/archive/gale.py
+++ b/ppa/archive/gale.py
@@ -1,4 +1,5 @@
 import os
+import json
 import logging
 import time
 
@@ -13,21 +14,21 @@ from ppa import __version__ as ppa_version
 logger = logging.getLogger(__name__)
 
 
-def get_local_ocr(item_id, page_num):
+def _get_local_ocr(item_id, page_num):
     """
     Get local OCR page text for specified page of a single Gale volume if available.
     This requires a base directory (specified by GALE_LOCAL_OCR) to be configured and
     assumes the following organization:
-        
+
         * Volume-level directories are organized in stub directories that correspond to
           every third number (e.g., CW0128905397 --> 193). So, a Gale volume's OCR data
           is located in the following directory: GALE_LOCAL_OCR / stub_dir / item_id
-        
+
         * A page's local ocr text file is named [volume id]_[page number]0.txt, where the
           page number is a 4-digit string (e.g., "0004"). So, for page number "0004" in
           volume CW0128905397, the local ocr file is named "CW0128905397_00040.txt".
 
-    Raises a FileNotFoundError if the local OCR page text does not exist. 
+    Raises a FileNotFoundError if the local OCR page text does not exist.
     """
     ocr_dir = getattr(settings, "GALE_LOCAL_OCR", None)
     if not ocr_dir:
@@ -39,6 +40,13 @@ def get_local_ocr(item_id, page_num):
     ocr_txt_fp = os.path.join(ocr_dir, stub_dir, item_id, f"{item_id}_{page_num}0.txt")
     with open(ocr_txt_fp) as reader:
         return reader.read()
+
+
+def get_local_ocr(item_id):
+    # error handling still TODO
+    ocr_dir = getattr(settings, "GALE_LOCAL_OCR", None)
+    with open(os.path.join(ocr_dir, f"{item_id}.json")) as ocrfile:
+        return json.load(ocrfile)
 
 
 class GaleAPIError(Exception):
@@ -218,6 +226,9 @@ class GaleAPI:
         if gale_record is None:
             gale_record = self.get_item(item_id)
 
+        # todo: needs try/catch error handling
+        local_ocr_text = get_local_ocr(item_id)
+
         # iterate through the pages in the response
         for page in gale_record["pageResponse"]["pages"]:
             page_number = page["pageNumber"]
@@ -225,12 +236,13 @@ class GaleAPI:
             # Fetch higher quality local OCR text if possible, with fallback to Gale
             # OCR. Set a tag to indicate the usage of local OCR text.
             tags = []
-            try: 
-                ocr_text = get_local_ocr(item_id, page_number)
-                tags = ["local_ocr"]
-            except FileNotFoundError as e:
+            ocr_text = local_ocr_text.get(page_number)
+            # TODO: set local ocr tag here
+            if not ocr_text and page_number not in local_ocr_text:
+                # if ocr text is unset and page number is not present,
+                # try getting the ocr from the gale api result
+                logger.warning(f"No local OCR for {item_id} {page_number}")
                 ocr_text = page.get("ocrText")  # some pages have no text
-                logger.warning(f'Local OCR not found for {item_id} {page_number}')
 
             info = {
                 "page_id": page_number,
--- a/ppa/archive/gale.py
+++ b/ppa/archive/gale.py
@@ -1,6 +1,6 @@
-import os
 import json
 import logging
+import pathlib
 import time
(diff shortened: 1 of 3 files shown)


laurejt:
Overall, this seems like a reasonable approach to me.
- Consider using `pathlib` instead of `glob` and `os.path`
- I'm not sure I understand the difference between .json vs .jsonl. In both cases you may need to worry about missing pages.
- Since the ocr code is all in ppa-nlp, it seems reasonable to also include the ocr-to-json script. In the future this script might also do additional text transformations.

ppa/archive/gale.py
```diff
@@ -218,19 +226,23 @@ def get_item_pages(self, item_id, gale_record=None):
         if gale_record is None:
             gale_record = self.get_item(item_id)
 
+        # todo: needs try/catch error handling
+        local_ocr_text = get_local_ocr(item_id)
```
laurejt, line 230: This can throw an error if the underlying file doesn't exist right?

ppa/archive/gale.py
```diff
@@ -218,19 +226,23 @@ def get_item_pages(self, item_id, gale_record=None):
...
-                tags = ["local_ocr"]
-            except FileNotFoundError as e:
+            ocr_text = local_ocr_text.get(page_number)
+            # TODO: set local ocr tag here
+            if not ocr_text and page_number not in local_ocr_text:
+                # if ocr text is unset and page number is not present,
+                # try getting the ocr from the gale api result
+                logger.warning(f"No local OCR for {item_id} {page_number}")
                 ocr_text = page.get("ocrText")  # some pages have no text
-                logger.warning(f'Local OCR not found for {item_id} {page_number}')
```
laurejt, line 233: Seems like this would be better as an if/else clause, since the only time the Gale text should be fetched is when the page number is not in `local_ocr_text`

ppa/archive/gale.py
rlskoeser: yes, I was thinking to add similar error handling to what we implemented the first time

ppa/archive/gale.py
rlskoeser: ah, good point... there are two cases here, aren't there? we could have the page number in our json file and have no text, or it might not be present at all (e.g., we missed it somehow)

rlskoeser:
@laurejt thanks for the quick review and useful feedback, I'll keep your recommendations in mind when I circle back to this.

jsonl vs json: I was thinking json lines so we could avoid having the whole file in memory, but that would mean we have to sort the pages before we generate the lines file (or when we load it), and then as we iterate we'd have to check if we were on the correct page - so logic is more complicated. I think these volumes are small enough that we can get by with one file per volume and just have the whole text in memory so we can look up contents by page number.

ppa/archive/gale.py
rlskoeser: Leaving this comment to remind you to confirm that I've done what you were suggesting.

I got a code complexity alert for this function - was able to reduce complexity slightly. If you see any easy wins, please suggest them.

ppa/archive/gale.py
```diff
@@ -13,32 +14,36 @@
...
+        raise ValueError(f"{item_id} is not a valid Gale item identifier")
 
-    stub_dir = item_id[::3][1:]  # Following conventions set in ppa-nlp
-    ocr_txt_fp = os.path.join(ocr_dir, stub_dir, item_id, f"{item_id}_{page_num}0.txt")
-    with open(ocr_txt_fp) as reader:
-        return reader.read()
+    # files are in stub directories; following conventions set in ppa-nlp
+    stub_dir = item_id[::3][1:]
+    ocr_dir = pathlib.Path(ocr_dir)
+    with (ocr_dir / stub_dir / f"{item_id}.json").open() as ocrfile:
```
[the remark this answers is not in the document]
rlskoeser: I thought my check that item id is formatted like a valid gale id would be enough here - what do you think @laurejt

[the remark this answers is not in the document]
laurejt: So, I think the thing it's trying to do is to ensure that there aren't any rogue `..` floating around... but we know this to be the case

laurejt, line 45: Why not just make the full path for the ocrfile?
```suggestion
    ocr_path = pathlib.Path(ocr_dir, stub_dir, f"{item_id}.json")
    with ocr_path.open() as ocrfile:
```

ppa/archive/gale.py
```diff
@@ -218,19 +223,34 @@ def get_item_pages(self, item_id, gale_record=None):
...
-                tags = ["local_ocr"]
-            except FileNotFoundError as e:
-                ocr_text = page.get("ocrText")  # some pages have no text
-                logger.warning(f'Local OCR not found for {item_id} {page_number}')
+            ocr_text = None
+            if local_ocr_text:
+                ocr_text = local_ocr_text.get(page_number)
+                # if we have content, set tag to indicate local ocr
+                if ocr_text:
+                    tags = ["local_ocr"]
```
laurejt, line 245: Should this be set if the resulting OCR for the page is blank?

ppa/archive/tests/test_gale.py
```diff
@@ -15,22 +16,32 @@
...
     item_id = "CB0123456789"
-    page_num = "0001"
-    content = "Testing...\n1\n2\n3"
+    content = {"0001": "Testing...\n1\n2\n3"}
     # Mock ocr files for testing
-    ocr_dir = tmp_path.joinpath("147", item_id)
-    ocr_dir.mkdir(parents=True)
-    ocr_file = ocr_dir.joinpath(f"{item_id}_{page_num}0.txt")
-    ocr_file.write_text(content)
+    ocr_dir = tmp_path / "147"
```
laurejt, line 21: No need to change the code, but the `/` notation for `join_path` messes with me (because it immediately makes me worry about file systems)

ppa/archive/tests/test_gale.py
```diff
@@ -15,22 +16,32 @@
...
+    ocr_dir.mkdir()
+    ocr_file = ocr_dir / f"{item_id}.json"
+    with ocr_file.open("w") as outfile:
```
laurejt, line 24: I don't know if we need to be defensive here, but the default encoding is platform dependent and not necessarily "utf-8 (although usually it will be).

```suggestion
    with ocr_file.open("w", encoding="utf-8") as outfile:
```

ppa/archive/gale.py
rlskoeser: nope, good catch

ppa/archive/tests/test_gale.py
rlskoeser: I tried to look to see if one of them was considered more pythonic or more idiomatic but didn't find much (didn't look very long). I like this syntax better because its terser.

ppa/archive/gale.py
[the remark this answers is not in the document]
rlskoeser: agreed. I'll dismiss it, then

ppa/archive/gale.py
rlskoeser: wait, maybe I misunderstood your question - that's what the logic is now, do you disagree with it?

laurejt: It seems like the tag should be set if the text is blank, because the empty string is a judgment of the local OCR (and not the Gale OCR)

laurejt (approved):

PR #691, opened 2024-11-04
Feature/upgrade python

laurejt:
Attempting to upgrade Python to 3.12 so that `ppa-nlp` can be used as a dependency


laurejt:
An upgrade to Python 3.12 is not possible because of `eulxml` which relies on `pkg_resources` which Python 3.12 has removed from the standard library (and to `setuptools`). That said, the use of `pkg_resources` is deprecated and `importlib`-based alternatives should be used instead ([source](https://setuptools.pypa.io/en/latest/pkg_resources.html)).

Per @rlskoeser's discussion with Emory, we should consider creating a hard fork of `eulxml` in the future.

For now, I will alternatively try adding 3.11 support to `ppa-nlp`.

PR #697, opened 2024-11-15
Increasing code coverage

laurejt:


blms:
@rlskoeser I added a couple of unit tests for things that Codecov flagged in my upgrades PR.

Regarding the PR description checkboxes, as far as I could tell, `page_index_data` is fully covered by unit tests. But it might be helpful to add testing for different expected functionality, just not strictly needed for coverage.

LMK if there's anything else I should do here before merging—were there management commands I should have coverage ignore?

rlskoeser:
@blms let's hold off on writing more unit tests.

These manage commands can be exempted from code coverage:
- check_hathi_excerpts (sort of experimental, we decided not to pursue but might want the code to write about the problem)
- import_clusters (one-time import, don't expect to repeat)

You want to do that and see where it gets us?

rlskoeser (approved):

blms:
@rlskoeser That improved coverage a bit! Still not 95%, but getting closer.

Perhaps I shouldn't have also added them to codeclimate exclude patterns, because now that tool is complaining about some other code 😬

blms:
@rlskoeser Great idea and it seems to be working well. It looks like there are about a combined 13 lines of tests that are not being executed. Should I handle, remove, or ignore before merging this?

rlskoeser:
> @rlskoeser Great idea and it seems to be working well. It looks like there are about a combined 13 lines of tests that are not being executed. Should I handle, remove, or ignore before merging this?

@blms please see if you can easily correct the skipped test lines (either fix or remove). Flag if anything starts taking too long.

rlskoeser (approved):

PR #699, opened 2024-11-18
Feature/improve hathi images

laurejt:


rlskoeser:
Closing since we decided not to pursue further development on this script.

PR #704, opened 2025-02-25
Identify quoted poems in TCP XML 

rlskoeser:
draft PR to get feedback on approach and recommendations for what and how much to revise; this is for https://github.com/Princeton-CDH/ppa-nlp/issues/145

changes included in this PR:
- revisions to the xml logic for eebo-tcp P4 TEI to also work with the P5 TEI used in the ecco-tcp content
- xml object for quoted poems, with logic for generating text chunked by `<pb>` tags (similar to page text logic used for EEBO-TCP)
- manage command to generate a CSV of the identified quoted poems (provisional format for review)

Here's a csv from the most recent run of the script, so you can see what the current output looks like and examples of excerpts that the code doesn't currently find in the ppa text
[tcp_quotedpoems.csv](https://github.com/user-attachments/files/18969073/tcp_quotedpoems.csv)

known issues / limitations:
- can't always identify start/end indices in ppa page text
   - whitespace differences in P5 / ecco content
   - footnote marker differences in both eebo and ecco
   - length of excerpt from xml doesn't always match the length of the excerpt on the ppa page (worse for ecco)
- non-latin text : could potentially use a regex for non-latin characters to find e.g. greek poetry in ecco (examples in the review output)

remaining work:
- improve start/end index logic and excerpt identification (and/or omit unfindable excerpts)
- map results to unlabeled excerpt dataclass for output


ppa/archive/management/commands/tcp_quotedpoems.py
```diff
@@ -0,0 +1,286 @@
...
+    on the page, if possible. Returns a tuple of start index, end index;
+    indices are None if not found.
+    """
+    start_index = end_index = None
+    # ecco-tcp includes long S that ecco-ocr renders as f
+    # for simplicity, replace in both texts before comparing
+    # (since this is a single-character replacement, it doesn't impact index)
+    _excerpt = excerpt.replace("ſ", "f")
+    _page_text = page_text.replace("ſ", "f")
+    # also long hyphens vs short in ecco; use ftfy ?
```
rlskoeser, line 267: does it make sense to use ftfy here to clean up both texts before comparing? does that include changes that will impact the character counts / indexes?

ppa/archive/management/commands/tcp_quotedpoems.py
```diff
@@ -0,0 +1,286 @@
...
+    of PPA, find and return the start and end indices of the excerpt
+    on the page, if possible. Returns a tuple of start index, end index;
+    indices are None if not found.
+    """
+    start_index = end_index = None
+    # ecco-tcp includes long S that ecco-ocr renders as f
+    # for simplicity, replace in both texts before comparing
+    # (since this is a single-character replacement, it doesn't impact index)
+    _excerpt = excerpt.replace("ſ", "f")
+    _page_text = page_text.replace("ſ", "f")
```
laurejt, line 266: Is it ever the case we might have an "s" be in one of these texts instead? I worry about introducing errors by converting long ess ("ſ") into "f"s

ppa/archive/management/commands/tcp_quotedpoems.py
laurejt: Unfortunately, it's very possible for `ftfy` to make transformations beyond the character-to-character level. The easiest example being if there are any ligatures.

Part of me wishes we could `ftfy` into the initial workflow for the PPA texts we're working with... Could we do this? Just as an assumption for all the ppa-nlp efforts at least.

ppa/archive/management/commands/tcp_quotedpoems.py
```diff
@@ -0,0 +1,286 @@
...
+    # also long hyphens vs short in ecco; use ftfy ?
+
+    try:
+        # best case scenario: the entire text matches exactly
+        start_index = _page_text.index(_excerpt)
+        end_index = start_index + len(excerpt)
```
laurejt, line 272: Is there any validation of the text matching between the XML and PPA texts? At the very least it could be worth adding a note if the match quality is poor enough. Some possible ways to compare:
- use a text similarity score (perhaps relying on something from `rapidfuzz`
- simple text match of a short prefix / suffix of each text

ppa/archive/management/commands/tcp_quotedpoems.py
```diff
@@ -0,0 +1,286 @@
...
+    except ValueError:
+        # some mismatches are due to the note marks;
+        # could try removing the note marks from the page text to find
+        # the content start and end?
```
laurejt, line 276: How much of an impact do note marks have? Removal could be worth it if it has a big enough impact, otherwise I'd table that improvement

ppa/archive/management/commands/tcp_quotedpoems.py
```diff
@@ -0,0 +1,286 @@
...
+
+        if not start_index:
+            lines = _excerpt.split("\n")
+            # if we have more than one line, try to get
+            # start and end based on  first and last line
+            if len(lines) > 1:
+                start_index, _ = get_excerpt_span(lines[0].strip(), page_text)
+                _, end_index = get_excerpt_span(lines[-1].strip(), page_text)
```
laurejt, line 284: Perhaps we could use rapidfuzz's `partial_ratio_alignment` to attempt to find the starting and/or ending line (or really just a starting or ending substring)? It's important that this search is on a short string.

ppa/archive/management/commands/tcp_quotedpoems.py
rlskoeser: I suspect both are possible... I chose this based on the mismatches I was seeing, which was the long ess being recognized as f in our local ocr. The google vision ocr does also include long ess characters as well, fwiw.

laurejt (commented):
There are two separate cases here:
- **EEBO-TCP** where the PPA texts and the extracted XML texts are to my knowledge the same; and therefore the excerpt indices between text will just match.
- **ECCO-TCP** where the PPA texts and extracted XML texts will often not be exact matches.

In this latter case, it seems like in this case we will need to perform a (hopefully) simple text alignment. We know that the XML excerpt will exist in some (hopefully) slightly altered form in the PPA text, so we should be able to find the PPA excerpt indices by performing a text alignment method. Perhaps `rapidfuzz`'s `partial_ratio_alignment` method could work? There may be other more suitable alternatives. We know the XML excerpt maps to some text span in the PPA text, we're just not sure of the exact indices.

Also, in terms of the output for the spreadsheet, it might be worth flagging when the excerpt spans are estimates or likely to be inaccurate.

ppa/archive/management/commands/tcp_quotedpoems.py
rlskoeser: I like the rapidfuzz text similarity score idea; could include that in the notes field

laurejt:
One additional thing: for notes we might want to add additional markers so we might know the context of a note when aggregation happens later one.

rlskoeser:
@laurejt thanks for the review and suggestions, these are helpful!

laurejt:
I just realized that we're not guaranteed to know the indices for EEBO. Is this caused primarily by excerpts occurring within note tags? In this case, it seems like there's two ways to deal with it:

1. If the text is dropped and not actually present in the PPA text corpus, drop the reference. I don't see a way of keeping it since it's not an excerpt present within the actual PPA text corpus itself.
2. If the text is within the excerpt, but it's less clear where. Fall back to text alignment, but this should hopefully be even easier since it should be an exact (or near it) substring of the PPA page text.

rlskoeser:
@laurejt revised to skip any excerpts where indices can't be found and not using alignment, per your advice, since it only affects ~50 excerpts out of 3 thousand some. Attaching output from the most recent run. 

[tcp_quotedpoems.csv](https://github.com/user-attachments/files/19438497/tcp_quotedpoems.csv)

ppa/archive/eebo_tcp.py
```diff
@@ -59,6 +60,24 @@ def parent_note(self, text):
 
         return within_note
 
+    def within_bibl(self, text):
+        """Check if a text element occurs within a BIBL element; if so,
+        return the BIBL element"""
```
laurejt, line 65: Is a BIBL element a bibliography element?

ppa/archive/management/commands/tcp_quotedpoems.py
```diff
@@ -271,10 +253,35 @@ def get_excerpt_span(excerpt, page_text):
...
+        )
+        # replace ſ with character set to match any of ſ, s, or f
+        regex_safe_excerpt = regex_safe_excerpt.replace("ſ", "[ſfs]")
+        # collapse any repeated whitespace to a single whitespace,
+        # then replace with a regex to match on any whitespace
+        regex_pattern = re.sub("\s+", " ", regex_safe_excerpt).replace(" ", r"\s*")
+        try:
+            text_re = re.compile(regex_pattern, flags=re.MULTILINE | re.UNICODE)
+        except SyntaxError:
+            # regex compilation could failed, e.g. due to unescaped special
```
laurejt, line 271:
```suggestion
            # regex compilation failed, e.g. due to unescaped special
```

laurejt (approved):
Looks good to me

ppa/archive/eebo_tcp.py
rlskoeser: yes

Issue #703, opened 2025-02-19
Upgrade the python, Django, wagtail stack for PPA website


jerielizabeth:
Currently python 3.11, django 5, wagtail 5.2

Tasks

- [ ] Upgrade python to 3.12
- [ ] Upgrade django to 5.1
- [ ] Upgrade to wagtail 6.3

blms:
I'm not seeing that behavior for title/subtitle matches only, locally, so I'm guessing the two are related (i.e. solr thinks there _are_ inner page hits, and those hits should be showing up there).

@rlskoeser Wondering if I should:
1. Create some data with full text OCR to test inner page hits locally—it wasn't immediately clear to me how to do that in the django admin, but I might be missing it.
2. Run the replicate playbook for PPA

rlskoeser:
@blms documenting what we discussed when we met today:
- replicating data from production and reindexing pages would be good to make sure this isn't a data issue
- I'll get you some full text content to work with so you can test this functionality locally

blms:
> Hits 2 and 3 just say "matching page" because they ostensibly only hit on the title/subtitle. But two issues 1) if there's a title/subtitle hit only, "matching page" should not appear. 2) in production, these works do have hits. Rhymed Hexameters has 1 hit, and The quantity, or measure of Latin syllables has 9.

This one appears to be an issue with the staging index, as reindexing those particular records has resolved it.
https://test-prosody.cdh.princeton.edu/archive/?query=pentameter
Since the reindex didn't complete, some records may still be broken in this way, but it doesn't appear to be a code issue.

> Gale image thumbnails not working; they display the "image not available" card ([e.g. here](https://test-prosody.cdh.princeton.edu/archive/CW0116738363/?query=rhyme)). This usually happens when the chron job fails to rotate the API encryption keys.

These appear to be working now, so maybe it's an intermittent issue related to the cron job's timing?

> CMS system won't let me upload a new SVG. (I can add an existing one, though.) I don't get an error message -- it just loads for a long time and then returns me to the "Choose a Document" screen. I'm pretty sure this behavior is different from production because I have successfully uploaded SVGs in the past.

> While testing the rich text options, I got the error message quoted below; I think I narrowed it down and discovered it occurs when code formatting occurs in the same block as other rich text. Note that I am not sure whether this behavior is also present on production! I may have found a separate bug...

[...]

PR #706, opened 2025-04-23
Python 3.12, Django 5.2, Wagtail 6.4 upgrades (#703)

blms:
## In this PR

Per #703:
- Upgrade Python to 3.12 (including `eulxml` to `neuxml`)
- Upgrade Django to 5.2
- Upgrade Wagtail to 6.4

### Upgrade notes

Not much to report here. The Wagtail upgrade was a bit different from S&Co (had to regenerate fixtures using the guide in dev notes, and didn't have to mess with rich text stuff). Django upgrades were mostly seamless since they were just minor version bumps. Python upgrade was mainly just `neuxml`.

### Other notes

Not sure why the unit test coverage is dipping so much. I do see two uncovered lines mentioned in the codecov report—[`HathiObject.mets_xml`](https://github.com/Princeton-CDH/ppa-django/blob/2b3126d5a34618bf73a2c9c8c81652685bcc22b4/ppa/archive/hathi.py#L349) and [`DownloadStats.log_error`](https://github.com/Princeton-CDH/ppa-django/blob/2b3126d5a34618bf73a2c9c8c81652685bcc22b4/ppa/archive/management/commands/hathi_images.py#L59)—should I go ahead and add tests for them?

- EDIT: Interesting, it claims that project coverage on the `develop` branch is actually _lower_, so maybe it's now trying to calculate coverage for some files it shouldn't?


ppa/archive/eebo_tcp.py
```diff
@@ -27,16 +27,16 @@ def short_id(volume_id):
...
+class TeiXmlObject(core.XmlObject):
     ROOT_NAMESPACES = {"t": TEI_NAMESPACE}
 
 
 class MixedText(TeiXmlObject):
     divider = "∣"
 
     # count following notes as a quick check to bail out of note detection logic
-    # (can't use eulxml boolean field here because it assumes string values for true/false)
-    has_notes = xmlmap.IntegerField("count(following::NOTE)")
```
rlskoeser, line 39: If I'd thought about how much the code was going to change based on the revision to the imports I might have changed my answer to how you changed this in neuxml!  (I recognize it's kind of late now).  This feels much less readable/obvious to me (although I'm also definitely biased after using the old syntax for so long).

LMK if you have any thoughts; how hard to support both with explicit / non-\*  imports for existing fields?

rlskoeser (approved):
Revisions look fine.

rlskoeser:
@blms regarding the code coverage - we've added some experimental manage commands that we didn't fully test, so the coverage has been slowly going down. I was thinking to go back and exempt some of those experimental scripts from code coverage, for things where we're not actually using or planning to use the functionality in production. If you see easy places to improve coverage, that would be helpful. 

It looks like there are some old branches that never got merged - some of which are stale and no longer relevant and can be removed, and others that might help the code coverage (I see one related to code coverage and another related to the hathi image script, which was an effort we abandoned - we ended up going a different direction for that). 

Do you have enough context to look briefly at the open branches and make some recommendations about which to drop and which might be worth merging?

blms:
@rlskoeser Here's what I can gather from the branches:

- [`feature/solr-stem-punc`](https://github.com/Princeton-CDH/ppa-django/compare/main...feature/solr-stem-punc)
   - Tiny change, seems like a planned feature / bugfix changing the way a certain field was analyzed and searched, recommend deleting the branch and mentioning the few changed lines in an issue if the work still needs to be done
- [`ecco2023`](https://github.com/Princeton-CDH/ppa-django/compare/main...ecco2023)
   - Looks like jupyter notebooks prototyping a plan to maybe import records from a new authority, seems like it would be a good feature but not sure if the jupyter notebooks should live here
- [`feature/solr-upgrade`](https://github.com/Princeton-CDH/ppa-django/compare/main...feature/solr-upgrade)
    - Probably not needed as it seems the solr upgrade ended up [happening separately](https://github.com/Princeton-CDH/ppa-django/pull/646), with some work from that branch
    - It looks like there was originally a plan to convert field references to implicit fields using suffixes in solr, so maybe that work should be preserved if it's still desired. I don't see any particular harm in leaving them explicit though.
- [`feature/upgrade-python`](https://github.com/Princeton-CDH/ppa-django/compare/main...feature/upgrade-python)
   - No longer needed
- [`feature/improve-code-coverage`](https://github.com/Princeton-CDH/ppa-django/compare/main...feature/improve-code-coverage)
   - It's a pretty small amount of coverage changes that we could easily roll into this PR or simply merge separately. Definitely worth keeping.
- [`feature/improve-hathi-images`](https://github.com/Princeton-CDH/ppa-django/compare/develop...feature/improve-hathi-images)
   - A pretty small number of changes to a management command, but if like you said this work is outdated, then definitely recommend deleting.

rlskoeser:
>  Here's what I can gather from the branches ...

@blms thank you so much for this helpful review! I've deleted most of the ones you flagged.

I'd like to get the improve-code-coverage changes merged (either separately or as part of this PR), and if it's not too difficult I'd be interested in getting the improve-hathi-images branch folded in. When I looked I saw, I think there are open PRs for both of these.

The priority should be getting your upgrade PR merged first (which I guess depends on the neuxml change, I'll look at that next); then fold in those other changes as seems reasonable to you.

Issue #559, opened 2023-11-07
Add a workflow for generating PDFs without developer assistance to use the new print styles/DOI fields feature in editorial


mnaydan:

blms:
Basic proposal:
- DocRaptor API key stored in Django settings
- A `/preview` endpoint in Django that is used to pass the contents of an `EditorialPage` document to DocRaptor's preview API
   - This will be responsible for processing Wagtail content into the needed format and applying styles as well
- A function, which may be called on `EditorialPage` save, that generates the finalized PDF w/ DocRaptor API
   - Should use factored-out logic (shared with preview endpoint) for handling content, styles and other needed settings
   - It will also save the file to configured Django storage and store a permalink to it on the `pdf` field

Proposal if DocRaptor's preview and final PDF generations are relatively fast and can be performed synchronously:
- A button in the Django admin that calls the `/preview` endpoint and live updates with the preview PDF's URL in the browser (maybe even just downloads it directly)
- A `EditorialPage` form boolean checkbox "Generate PDF" (doesn't need to be a DB field) that, on save, will generate the finalized PDF, upload it to storage, and store its url on the `pdf` field

Proposal if PDF generation is slow and needs to be done asynchronously:
- A task queue (or cron?) job that checks for completion of the job; for preview, will need to store the URL somewhere on the `EditorialPage` model, so that users can check back when it's ready

rlskoeser:
@blms thank for this! This sounds promising and I appreciate you thinking through the options.

I forgot to tell you, in the past we've deposited and linked to the PDFs in a system like Zenodo. But I guess that doesn't preclude saving it as a Django file if that makes the implementation easier.

Do you mind signing up for a free API key for testing & development purposes, so you can time how long it takes to generate a PDF? Last time I used it, you can generate unlimited test/watermarked versions, the 5/month limit is for the non-test copies.  

I expect the longer articles with more images will take longer; this piece of mine is probably a good test case for that https://prosody.princeton.edu/editorial/2020/01/visualizing-collections/  If you can test that and then one of the shorter ones, we ought to have a good sense of time range to make a decision about implementation.

PR #708, opened 2025-05-15
PDF generation in Wagtail (#559)

blms:
**Associated Issue(s):** #559

### Notes

- My approach changed somewhat as I learned more about Wagtail, and to some extent DocRaptor.
   - For example, you can't just get rendered HTML/CSS easily out of Wagtail as it would appear published, without new templates; a `Page` form panel isn't by default aware of its own published URL; and there is some [special configuration](https://docs.wagtail.org/en/stable/extending/extending_client_side.html#extending-with-stimulus) to get custom JS running inside of it (it uses Stimulus!).
- Had to make some tweaks to the webpack config and npm script to get the JS dev server working. Looks like it hadn't been updated for Webpack 4 yet.
- Generation is near-instant, so async JS without persistence is totally fine.
- I ended up not saving it to Django storage or anything since you mentioned Zenodo; seems fine to me to just let the user download the PDF and then host it themselves.
- Leaving this as a draft due to the below issues, and because it's still pretty experimental so I haven't added unit tests.

### Issues

- I can't really test locally that it works to pass the Wagtail url from `get_url()` to DocRaptor, since that's on `localhost` and contains a lot of internal links to local content (css, js, images, etc). But I tested just manually passing the url `https://prosody.princeton.edu/editorial/2020/01/visualizing-collections/` and it worked fine.
   - ... though, it did have some gray vertical bars on the left and right, and the font size is rather large. Should I make any changes to the print layout CSS? [Here's what the output looks like](https://github.com/user-attachments/files/20233893/preview.1.pdf)

   - This also means we won't be able to test it on the test site unless we allow non-firewall access to it temporarily. (DocRaptor does support passing rendered HTML as `document_content`, but I assume that will still link to CSS/JS/images on the test site, which will cause permissions errors in DocRaptor.)
- It seems that the [Wagtail API](https://docs.wagtail.org/en/stable/advanced_topics/headless.html#page-preview) doesn't give us a way to programmatically request a preview/draft version of the page. So, out of the box we can only really send DocRaptor the published page URL.
   - Even if we had a preview URL, I'm not sure we could pass that along to DocRaptor; at least, we'd need to pass along the fully rendered HTML to avoid permissions issues. Even then, I'm not 100% sure it would work, and I'd need to test on a live server. (that isn't firewalled)
   - That said, it may be possible to generate a preview using [wagtail-headless-preview](https://github.com/torchbox/wagtail-headless-preview) if we want to go that route.
   - My current solution is just to prevent the user from accessing DocRaptor at all if the page is still a draft or has unpublished changes. (The latter is so that people don't save some changes without publishing, and then get confused when they aren't reflected in the PDF.)

Some other possible options:
- Somehow passing ONLY the article content to DocRaptor, instead of the entire rendered page… then we could fetch the contents of the local CSS bundle and include the whole thing in `<style>` tags. But the problem is that any inline images won't work if they're still relative, referencing localhost, or referencing the firewalled URL. Maybe that's acceptable for test environments.
- Requiring the user to supply the URL for DocRaptor, similar to how the startwords script works, and prompting them in help text to supply the published production URL.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
         "url": "https://github.com/Princeton-CDH/ppa-django.git"
     },
     "scripts": {
-        "dev": "webpack serve --progress --hot --inline",
+        "dev": "webpack serve --progress --hot",
         "dev:semantic": "gulp --gulpfile sitemedia/semantic/gulpfile.js",
         "start": "npm run dev",
         "test": "npm run test:unit && npm run test:a11y",
--- a/ppa/editorial/models.py
+++ b/ppa/editorial/models.py
@@ -1,9 +1,10 @@
 from datetime import date
 
+from django.conf import settings
 from django.core.validators import RegexValidator
 from django.db import models
 from django.http import Http404
-from wagtail.admin.panels import FieldPanel
+from wagtail.admin.panels import FieldPanel, Panel
 from wagtail.fields import RichTextField, StreamField
 from wagtail.models import Page
 from wagtail.snippets.blocks import SnippetChooserBlock
@@ -89,6 +90,33 @@ validate_doi = RegexValidator(
 )
 
 
+class GeneratePdfPanel(Panel):
+    """Panel for asynchronous PDF generation in JavaScript; must be a
+    Panel subclass to have access to EditoralPage's instance URL"""
+
+    class BoundPanel(Panel.BoundPanel):
+        template_name = "wagtailadmin/panels/pdf_panel.html"
+
+        def get_context_data(self, parent_context=None):
+            """Override to insert live page URL and docraptor API key"""
+            context = super().get_context_data(parent_context)
+            url = self.instance.get_url()
+            # disable if unpublished, or has unpublished changes
+            if (
+                not url
+                or not self.instance.live
+                or self.instance.has_unpublished_changes
+            ):
+                url = ""
+            context.update(
+                {
+                    "url": url,
+                    "docraptor_api_key": getattr(settings, "DOCRAPTOR_API_KEY", ""),
+                }
+            )
+            return context
+
+
 class EditorialPage(Page, PagePreviewDescriptionMixin):
     """Editorial page, for scholarly, educational, or other essay-like
     content related to the site"""
@@ -125,6 +153,7 @@ class EditorialPage(Page, PagePreviewDescriptionMixin):
         FieldPanel("editors"),
         FieldPanel("doi"),
         FieldPanel("pdf"),
+        GeneratePdfPanel(),  # use custom panel for PDF generation
         FieldPanel("body"),
     ]
 
--- a/ppa/editorial/templates/wagtailadmin/panels/pdf_panel.html
+++ b/ppa/editorial/templates/wagtailadmin/panels/pdf_panel.html
@@ -1,0 +1,37 @@
+{% if docraptor_api_key %}
+    <div
+        data-controller="pdf"
+        data-pdf-url-value="{{ url }}"
+        data-pdf-apikey-value="{{ docraptor_api_key }}"
+    >
+        <h3 data-pdf-target="heading">Generate a PDF copy of this page:</h3>
+        <div>
+            <button
+                class="button button-secondary"
+                data-pdf-target="preview"
+                data-action="pdf#generatePdf"
+                type="button"
+            >
+                Generate preview PDF
+            </button>
+            <span data-pdf-target="downloadPreview"></span>
+            <p class="help">
+                Unlimited previews allowed, but will be watermarked.
+            </p>
+        </div>
+        <div>
+            <button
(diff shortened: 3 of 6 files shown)


rlskoeser:
@blms I've scanned through the code changes and they all seem reasonable (haven't tested anything or looked too closely). 
This is very exciting progress, and your approach seems good — I'm glad to hear it's fast enough, that simplifies things. Appreciate the thoughtfulness.

Seems more important to answer your questions quickly.

> * I ended up not saving it to Django storage or anything since you mentioned Zenodo; seems fine to me to just let the user download the PDF and then host it themselves.

This sounds like a good solution to me, appreciate the simplicity.

> ### Questions
> 
> 1. Should we just require the user to supply a URL, similar to how the startwords script works, and prompt them in help text to supply the published production URL?

Yes, let's use production URLs - that's simpler, and good enough for the immediate needs. When this goes through acceptance testing, can we test the code functionality but use the production site URLs? What limits do you / should we apply to the URLs we support converting to PDF?

Can you also document and add notes for local development / testing?  I remember docraptor documentation included some suggestions, IIRC there's a way to spin up a temporary proxy or tunnel that allows you to generate PDFs from a local site.

> 2. The previews I generated from the production site have some gray vertical bars on the left and right, and the font size is rather large. Should I make any changes to the print layout CSS? [Here's what the output looks like](https://github.com/user-attachments/files/20233893/preview.1.pdf)

Yes - but let's make it a separate issue and track the work separately, if that makes sense to you? (I know the work is related).  I generated a few PDFs before, you can see what they look like on the ones generated before, see the PDF link here: https://prosody.princeton.edu/editorial/2019/03/coventry-patmore-and-problem-excerpts/

I would guess that you're using a different version of DocRaptor than I was, but updating to a newer version seems good.  The gray background looks like the site background isn't getting overridden for page content for some reason.  I hadn't initially intended for the PPA header to be centered on the page, I think it was a result of some interaction with DocRaptor styles (I was adapting styles from Startwords) - but I do like how putting it there matches the site.

blms:
@rlskoeser The tunneling worked! The one DocRaptor mentions didn't work, because it always redirects to an interstitial page that requires a click to pass through, at least on the free tier. But I mentioned one that works in the notes in 51fa230.

Now we will just need to figure out how to test in QA. I assume whitelisting DocRaptor in the firewall will work, though I couldn't find anywhere in their docs where they list hostnames/IPs they might use.

And I created the new issue: #708 Though, when I used the tunneling I didn't actually see any issues with the CSS, so maybe this is something that's already fixed in QA and can be closed once we're able to test DocRaptor there.

rlskoeser:
@blms I was looking at DocRaptor API docs to see if there's any programmatic way to check on api limits and usage - didn't find any, but I found this option for referrer-based documents:

https://docraptor.com/documentation/api/referrer_based_documents

I don't remember seeing this before! Sorry I was not aware of it. Would you test and see if you can get this to work and if you can generate a test document with a referrer link? 

I know it would mean throwing away some (most?) of your work in this PR, but in the long run it seems like it would be much easier to maintain.  My thought would be to add to the public template a block that only logged in users could see, to download a test pdf and final pdf (probably language similar to what you have now).  Could you take a look and give me your recommendation?

blms:
@rlskoeser I'm not able to get it to work on my local instance using a tunnel. Not sure if that means it won't work at all. At first the link on PPA didn't send the `HTTP_REFERRER` header at all, so the API endpoint errored; when I included a `meta` tag on the editorial page to force a more lenient referrer policy, I got redirected to the DocRaptor document logs, with no indication that it actually received my request or attempted to create a PDF from the page.

Under the hood it seems like it likely would do the same thing as the normal "request with API key" option, and visit the webpage included int he referrer header itself. The only difference seems to be that it uses matching with the URL whitelist you supply, instead of an API key, for authentication. So it would still require the same dev overhead with setting up a tunnel etc, and would run into the same kind of issues with the test server—it's not sending the rendered HTML from your machine.

So my recommendation would be to stick with the solution here unless there is a strong reason why the link should be on the page itself rather than in the editor.

rlskoeser:
@blms thanks for testing and documenting. I agree with your recommendation.

blms:
@rlskoeser looks like I wasn't able to use the `template_globals` context processor in Wagtail, but I just included it in the panel's context.

Should we hold off on merging this one until PPA smoke testing is done, or is it ok to merge this and prep testing notes since that may not have started yet?

rlskoeser:
> @rlskoeser looks like I wasn't able to use the `template_globals` context processor in Wagtail, but I just included it in the panel's context.
> 
> Should we hold off on merging this one until PPA smoke testing is done, or is it ok to merge this and prep testing notes since that may not have started yet?

Let's go ahead and merge it - should be more efficient if this can be tested with other new features, and it doesn't seem like it should cause any problems with wagtail.

PR #722, opened 2025-06-02
Set DocRaptor key in Wagtail, not Django settings (#559)

blms:
**Associated Issue(s):** #559

### Changes in this PR

- Add a new Wagtail settings module for DocRaptor and remove local settings for it
- Update context processors and template snippet to reference Wagtail setting


rlskoeser (approved):
Great!

Issue #718, opened 2025-05-30
As a content editor, I want to add tags to editorial essays so I can categorize and group essays for discovery.


rlskoeser:
Enable tagging for wagtail editorial pages — use taggit or whatever the standard solution is.

mnaydan:
Unable to test final item in checklist. Upon trying to save as draft, I get the following error message:

<img width="493" alt="Image" src="https://github.com/user-attachments/assets/df940b24-7713-4429-8e10-696008c7b57a" />

Not sure what's going on. I double checked that I'm connected to VPN.

mnaydan:
Works well! Firewall was the issue. Nice detective work.

PR #733, opened 2025-06-11
Add tags to EditorialPage in Wagtail (#718)

blms:
**Associated Issue(s):** #718 

### Changes in this PR

- Adds wagtail tags to EditorialPage following [this example](https://docs.wagtail.org/en/latest/advanced_topics/tags.html)

### Notes

- There doesn't seem to be any filtering or searching on tags in the CMS out of the box. We could implement a `PageListingViewSet` override if that's a requirement.


rlskoeser:
> @rlskoeser Looks like you can [add a snippet viewset](https://docs.wagtail.org/en/latest/advanced_topics/tags.html#managing-tags-as-snippets) to view the list and add/edit/delete, but not sure it includes counts out of the box. Should I add that to this PR?

Yes, please add - basic functionality is fine for a start!

Issue #547, opened 2023-08-25
Help text "Displaying x digitized works or clusters of works" reverts to "Displaying x digitized works" after search


mnaydan:
### Testing notes (QA)

On the [test site search](https://test-prosody.cdh.princeton.edu/archive):
- [x] The "clusters of works" text should not disappear after keyword searching

---

The help text on the archive page beneath the search box ("Displaying x digitized works or clusters of works") should persist even after search to indicate to the user that clusters of search may appear in the results; however, the text reverts to the original text prior to cluster implementation after search.

mnaydan:
Help text now persists as expected after keyword searching, filtering, and navigation in and out of hits.

PR #734, opened 2025-06-11
Include cluster result text in results_list (#547)

blms:
**Associated Issue(s):** #547

### Changes in this PR

- Update `results_list.html` template, which replaces part of `search_results.html` via async JS, to include its "or clusters of works" text
- Use count pluralization on the word "cluster" too, in case the result is 1 cluster of works


rlskoeser (approved):
ah, thank you, such a small fix once you look in the right place

Issue #300, opened 2019-05-08
Decrease responsiveness of title and author search fields only


mnaydan:
### Testing notes (QA)

On the [test site](https://test-prosody.cdh.princeton.edu/archive):
- [x] When typing in the keyword, title, and author fields, they should all have a reasonable delay before the loading gif appears.

---

Nick and I discussed decreasing the responsiveness of the title and author search fields because right now the speed of the search exceeds the speed at which a user can reasonably type, causing the loading gif to appear and reappear in a jumpy way as it searches partial queries. (The speed for keyword search is perfect because it needs to comb through more data to yield the results.)

mnaydan:
I just retested this -- and it looks like it is fixed for author field but title field is still jumpy. It might be good to fix before we deem the site stable.

mnaydan:
Passed! Tested compared to production and it's much better on test -- if you're typing reasonably slowly, it will wait rather than tell you there's 0 results and jumping around.

PR #737, opened 2025-06-11
Increase debounce time to 750ms (#300)

blms:
**Associated Issue(s):** #300

### Changes in this PR

- Bump debounce time for form fields from 500 to 750ms

### Notes

- I went with 750ms because while 500ms seems reasonable to me, 1000ms means I actually have to wait for results to start showing up; it feels sluggish. But this is subjective and really depends on the person's typing speed.
- The original issue notes that it seems fixed for some fields but not others, but the debounce time is the same for all three, and theoretically this should be independent of data load times because the loading indicator should show up regardless. But maybe it's just a quirk of how people are using it?


rlskoeser (approved):
Thanks, trust your testing and sense on this one.

Issue #724, opened 2025-06-02
Simplify fomantic LESS/JS build process


blms:
> Fomantic UI: The Gulp build process involved in `fomantic-ui` is a potential area of concern.
>    - Gulp is a fairly outdated build tool that developers may not be familiar with. However, it’s required to build the full theme for `fomantic-ui`.
>    - It requires including a lot of vendor code, in particular vendor JS, in PPA source code, which must also be updated every time the package is updated.
>    - Also, right now `npm audit` reports a couple of `fomantic-ui` dependencies as security holes, so it is fairly important to make it easier to update.
>    - Recommendation: Use prebuilt JS from the existing `fomantic-ui` package, add `fomantic-ui-less`, which is only the LESS source code and is capable of reading PPA’s theme overrides, and finally use webpack `less-loader` to compile it. That way we can preserve our theming without Gulp and unnecessary vendor JS.

 _Originally posted by @blms in [#709](https://github.com/Princeton-CDH/ppa-django/issues/709#issuecomment-2931921245)_

---

### dev notes

- [ ] Use prebuilt JS from the existing `fomantic-ui` package instead of vendor code in the repo
- [ ] add `fomantic-ui-less`
- [ ] use webpack `less-loader` to compile the LESS, ensuring theme overrides also included

PR #743, opened 2025-06-25
Replace Fomantic gulp build with webpack build (#724)

blms:
**Associated Issue(s):** #724

### Changes in this PR

- Replace fomantic gulp build with JS import from `fomantic-ui` package, LESS import from `fomantic-ui-less`, and webpack config to handle LESS overrides
- Remove vendor JS and gulp config (that's what most of the changed files are!), as well as gulp build process scripts
   - Update readme to reflect changes
   -  Remove gulp from the project entirely
- Update included vendor LESS definitions to match installed fomantic-ui version
   - Small tweak to `base.html` for more recent fomantic-ui `pusher` changes
- Add dev notes about how to upgrade fomantic-ui if needed (though it seems unlikely much will change)

### Notes

- In reviewing, you can ignore all the `sitemedia/semantic/src/themes/default` changes, since that is the remaining required vendor code. The changes are the upgrade from an older fomantic version to v2.9.4.
   - It's still required to include the `default` theme vendor LESS code in `themes/` due to how fomantic uses our local overrides inside the theme, so I've explained how to update that in the dev notes.
- Tested locally by comparing all styles and fonts to the production instance, as well as testing semantic-ui JS functions like `popup()`.
- Done using these guides (in addition to various Webpack 5 resources to adapt some of the outdated configs):
    - https://marekurbanowicz.medium.com/how-to-customize-fomantic-ui-with-less-and-webpack-applicable-to-semantic-ui-too-fbf98a74506c
    - https://neekey.net/2016/12/09/integrate-react-webpack-with-semantic-ui-and-theming/
    - https://dev.to/kudakwashegore/how-to-customize-semantic-ui-with-react-minicssextractplugin-less-and-webpack-4-498j
- The webpack build now takes a pretty long time, as it has to process all the fomantic UI code in webpack instead of gulp.
    - It's about 50-70 seconds for me locally. Not sure how long the gulp process took by comparison.
    - This also impacts hot reloading with webpack dev server.
- Just noticed, it appears that the `build:qa` and `build:prod` scripts are identical even though the readme claims they're different in terms of source maps. Should I investigate?


rlskoeser (approved):
Thanks for figuring this out and documenting the upgrade process.  The slower build is tolerable.

I pulled this branch locally and was able to successfully run the build. Anything else you think I should check?

I don't think we care about source maps at this point, since we're not planning to do further development. Let's drop the qa build everywhere. (Will likely require a slight tweak to the cdh-ansible staging deploy also).

I noticed we're still on node 18; should we update to node 20?

blms:
@rlskoeser Node 20 upgrade went smoothly for me locally. Opened a PR on cdh ansible with changes: https://github.com/Princeton-CDH/cdh-ansible/pull/276

(Waiting to merge this one until we resolve the unit tests issue https://github.com/Princeton-CDH/ppa-django/pull/744#issuecomment-3020920075)

blms:
Thinking it might make sense to hold off on this one until the remaining "needs attention" issues pass testing and we do a prod deploy. Seems wise not to introduce build and ansible changes until after everything in prod passes testing

Issue #725, opened 2025-06-02
Replace deprecated calls in JS/TS code


blms:
### dev notes

- [ ] [`keyCode` in JS](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)
- [ ] [`click()` in jQuery](https://api.jquery.com/click-shorthand/)
- [ ] [`pluck()` in RxJS](https://rxjs.dev/api/index/function/pluck) 

 _Originally posted by @blms in [#709](https://github.com/Princeton-CDH/ppa-django/issues/709#issuecomment-2931921245)_

PR #744, opened 2025-06-30
Replace deprecated JS calls (#725)

blms:
**Associated Issue(s):** #725

### Changes in this PR

- Replace the deprecated JS, RxJS, and jQuery calls

### Notes

- All JS/TS unit tests are passing locally


rlskoeser (approved):
JS code looks much better, thanks for updating.

rlskoeser:
OK, for the short-term fix let's pin lxml < 6.0 so that the PPA updates and testing can proceed.  We have a plan now and a start on the longer-term solution.

Issue #749, opened 2025-07-02
Improve Zotero citation workflow and performance


tanhaow:
## User Story
**As a researcher using Zotero,** I want to easily & quickly save bibliographic metadata for works (books, excerpt, etc.) in PPA to Zotero library **so that** I can properly cite them in my academic work.

**As a developer maintaining the PPA,** I want a simplified metadata workflow that uses our existing database fields **so that** I can reduce external api calls and improve performance.

## Problem
Our current unAPI metadata service has several issues:
- Complex MARC processing with external API calls to HathiTrust
- File-based pairtree storage for cached Gale MARC records  
- Redundant data fetching when we already have all citation data in our database
- Current code only handles books, not excerpts #399 
- Unnecessary binary metadata downloads
- Current links saved to Zotero are external links to HathiTrust or Gale, not PPA #404

## Solution
Replace the MARC-centric workflow with a Dublin Core-based approach that uses our existing `DigitizedWork` model fields in database.

mnaydan:
@tanhaow this looks good! I tested save a single item & multiple items with full work, excerpt, and article & HathiTrust, Gale, and EEBO-TCP. All of the key metadata appears present and correct.

The only thing I noticed was that one of the items I saved included a "Context Object" in the "Extra" field. The other items were not doing this, so it might be an outlier, but I wanted to flag quickly before closing.

<img width="819" height="527" alt="Image" src="https://github.com/user-attachments/assets/d47a9fdd-35f0-49af-a6d3-21037895bb42" />

PR #750, opened 2025-07-10
Remove unAPI and MARC, switch to COinS for Zotero metadata, and clean up related code

tanhaow:
**What this does**
- This PR completes the work in issue #749 by replacing the **MARC-based unAPI service** with **COinS metadata** that we generate directly from the `DigitizedWork` fields. 
- It also removes the `ppa.unapi` Django app and everything that depended on it. 
- Also, there are a few lines in models.py automatically reformatted by Ruff.


ppa/archive/templates/archive/digitizedwork_detail.html
```diff
@@ -13,12 +13,22 @@
 <meta name="twitter:label1" content="Publication Date" />
 <meta name="twitter:data1" content="{{ object.pub_date }}" />
 <meta name="twitter:label2" content="Author" />
-<meta name="twitter:data2" content="{{ object.author }}" />
+<meta name="twitter:dçata2" content="{{ object.author }}" />
```
rlskoeser, line 16: This change looks like a straw character was accidentally introduced

ppa/archive/templatetags/ppa_tags.py
```diff
@@ -113,3 +115,29 @@ def esc(x):
...
+
+
+@register.filter
+def last_page(page_range):
+    """Extract last page number from a page range like '25-30' or '25'"""
+    if not page_range:
+        return ""
+    parts = page_range.split("-")
+    return parts[-1].strip() if len(parts) > 1 else parts[0].strip()
+
```
rlskoeser, line 135: Are these used anywhere?

ppa/archive/templates/archive/snippets/search_result.html
```diff
@@ -1,11 +1,24 @@
...
 <li class="item {{ item.work_type }}">
-    {% if item.work_type == "full-work" %} {# unapi metadata only available for full works #}
-    <abbr class="unapi-id" title="{{ item.group_id }}" aria-hidden="true"></abbr>
+
+    {# Generate URL once for use in both COinS and main link #}
+    {% if item.first_page %}
+        {% url 'archive:detail' source_id=item.source_id start_page=item.first_page as detail_url %}
+    {% else %}
+        {% url 'archive:detail' item.source_id as detail_url %}
+    {% endif %}
```
rlskoeser, line 10: nice; took me a moment to remember why these differ, suggesting a comment to as a reminder
```suggestion
    {# Generate URL once for use in both COinS and main link #}
    {% if item.first_page %} {# generating detail url for excerpts requires specifying start page #}
        {% url 'archive:detail' source_id=item.source_id start_page=item.first_page as detail_url %}
    {% else %}
        {% url 'archive:detail' item.source_id as detail_url %}
    {% endif %}
```

ppa/archive/templatetags/ppa_tags.py
```diff
@@ -113,3 +115,29 @@ def esc(x):
...
+
+@register.simple_tag(takes_context=True)
+def absolute_url(context, relative_url):
```
rlskoeser, line 138: We already have [django-fullurl](https://github.com/Flimm/django-fullurl) as a dependency (used for a similar need in the open search header), so let's use the `fullurl` template tag instead of adding our own.

rlskoeser (changes requested):
Thank you for figuring out how to get COinS to work for PPA content!

The removal of all the unAPI-related code looks good.

I'm concerned about the duplication of COinS logic for the different work types and across the two output modes. Since this can't be a class method, I suggest two custom template tags: one to generate a dictionary of fields that handles the work/excerpt logic and also any differences between DigitizedWork and Solr results, and a second that turns the dictionary into the COinS span with escaped and combined fields.  This is just a suggestion, you are welcome to solve it a different way! The main goals for refactoring are to reduce duplication, improve readability, and make the logic more directly testable.

If I'm thinking about it correctly, once you have the tags created, then adding the COinS to the template could look something like this:
```
{{ coins_data item | coins_encode }}
```
(These names are arbitrary, just as an example, don't feel like you need to use them.)

I've commented on a few other small changes that are not needed or have alternative solutions, hopefully those will be simple to resolve.

rlskoeser:
> Also, there are a few lines in models.py automatically reformatted by Ruff.

I forgot to ask about this. Did you install the pre-commit hook? I see we have both black and ruff (which we probably shouldn't), but I thought we ran it on everything when we set it up.

tanhaow:
Again, thanks for your careful review! I will revise based on your suggestions.
Thanks for reminding me the pre-commit hook. I think I forgot to run it this time.

ppa/pages/templates/pages/home_page.html
```diff
@@ -1,6 +1,6 @@
 {% extends 'base.html' %}
 {% load render_bundle from webpack_loader %}
-{% load static humanize wagtailcore_tags %}
+{% load static humanize wagtailcore_tags ppa_tags %}
```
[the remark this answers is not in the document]
tanhaow: Same, `dict_item` filter from ppa_tags is used

ppa/editorial/templates/editorial/editorial_index_page.html
```diff
@@ -1,5 +1,5 @@
 {% extends 'pages/content_page.html' %}
-{% load wagtailcore_tags static %}
+{% load wagtailcore_tags static ppa_tags %}
```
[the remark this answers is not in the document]
tanhaow: `querystring_replace` tag from ppa_tags is used in this page

blms:
@rlskoeser Yes! It's fixed in `develop` by the commit 83a9a604362eabab5cc5254576f8ffcfdea9e0a0.

tanhaow:
> > Noticed there are unit test failures in the PR check, but they are completely unrelated to the COinS metadata changes we made. They're failing due to a network connectivity issue in the CI environment.
> 
> Yes - @blms do we have a fix for this yet? Is there a change in `develop` that @tanhaow needs to pull into this branch?

Apart from this, I've finished revising the code based on the earlier comments. Now the CoinS logic has been refactored into using custom template tags with dictionaries to construct and encode - which is indeed more neat and structured.

tanhaow:
@blms Thanks, Ben. Now the CI issue has been resolved. 

However, I still have two complaints from the `codeclimate` check. One of the complaints is related to the upgrade issue with `codeclimate` itself, which says Code Climate has been replaced by Qlty, and it will be discontinued soon. The other complaint mentions 4 issues that need to be fixed. However, when I go to the ‘view details’ section and run the upgraded Qlty check, it shows zero blocking issues. Therefore, I’m not sure what the 4 issues are. Given that codeclimate has been migrated to Qlty, do we still need this codeclimate check? @rlskoeser

rlskoeser:
> Given that codeclimate has been migrated to Qlty, do we still need this codeclimate check? @rlskoeser

No, we don't - I couldn't figure out how to remove it when I looked quickly last week! I'm not even sure if we want Qlty - feel free to ignore them both.

rlskoeser (commented):
As I mentioned in person, the code is more complicated than I expected it to be - I know part of this is because you have to balance two different types of objects (Solr Result and DigitizedWork model) as well as the different types of work for citation.  I think there are several places that can be simplified, and I'm hoping we can take better advantage of existing model or Solr functionality. I've commented on a few things, and we can look at this code and the ppa application code more together.

A lot of the code in the `ppa_tags` code is not strictly template code; if we do still need this many different functions to generate the COinS data then perhaps we should split it out into a new file, and import the necessary methods for use in the template tag.

ppa/archive/models.py
```diff
@@ -1014,7 +1057,11 @@ def count_pages(self, ptree_client=None):
             # some aggregate packages retrieved from Data API
             # include jp2 and xml files as well as txt; only count text
             page_count = len(
-                [filename for filename in ht_zip.namelist() if filename.endswith(".txt")]
+                [
```
tanhaow, line 1060: here is just format change done by Ruff

ppa/archive/templatetags/ppa_tags.py
```diff
@@ -113,3 +116,165 @@ def esc(x):
...
+def _convert_to_coins_format(existing_data):
+    """Return fields converted to COinS format."""
+    fields = {}
+    if "au" in existing_data:
+        fields["rft.au"] = existing_data["au"]
+    if "date" in existing_data:
+        fields["rft.date"] = existing_data["date"]
+    if "pub" in existing_data:
+        fields["rft.pub"] = existing_data["pub"]
+    return fields
```
rlskoeser, line 234: Do all the data fields get this 'rft' prefix? I like the approach you're taking here and would suggest using it everywhere - use the unprefixed field names everywhere and then add them all at the end (maybe as part of filtering out any nulls).

Maybe something like this:
```suggestion
   return {f"rft.{key}": val for key, val in existing_data.items() if val is not None}
```

rlskoeser (approved):
Revisions look great!  

Did you figure out what the error was that didn't cause a problem in your unit tests? Would be nice if you could test for that, whatever it was. 

I think we're in good shape to proceed with merging this and getting it on staging for acceptance testing.

rlskoeser:
> Yes—I tracked the gap to our forgiving Mocks: they happily returned values for missing Solr fields, so aliasing failures slipped past the old tests.

Ah, good work figuring that out! Yes, sometimes Mocks are too flexible.

rlskoeser:
@tanhaow I just noticed there's a test coverage error - we configured that to 100% because it usually means there's an error (e.g. duplicated test method name or nested assertion). It looks like it is related to your change to the mock — please resolve before you merge; let me know if you have trouble accessing or want any advice how to fix it.

https://app.codecov.io/gh/Princeton-CDH/ppa-django/pull/750?dropdown=coverage&src=pr&el=h1

Issue #754, opened 2025-07-21
fix indexing bug


mnaydan:
## Development notes
- [x] revise logging to exclude response content (only report 500 error and the url; possibly report item id separately)
- [x] see if we can determine whether retry succeeds (maybe info/warn logging based on retry count?)

## Testing notes

Can't be tested directly, but we should index all gale pages:

```console
python manage.py index_pages --gale
```

Expected behavior: if we get a 500 error, it should be logged and the item should be retried, but even if it fails the script should not crash.

tanhaow:
Revised this logging issue in #761

rlskoeser:
merged into develop in #761; included in release 3.15.1 and ready to go out next week

PR #755, opened 2025-07-22
Catch 500 error from Gale API and treat as a retry

rlskoeser:
**Associated Issue:** #754

### Changes in this PR

- Update Gale API code to catch 500 errors; log the error and retry. At a minimum this should allow us to complete the indexing without crashing when a 500 error happens, but hopefully the retry will give the server a chance to recover.
- Modified the unit test to confirm the new behavior

### Notes

- I'm not sure yet if this will resolve the indexing problem; I couldn't duplicate the problem in development. Thinking to try deploying this branch to staging and test indexing there to see if I can get another 500 error to happen.
- I tried to add a unit test check for the logging behavior, but I couldn't get the pytest `caplog` fixture to work in the unit test class structure we're using here

### Reviewer Checklist

- [x] Review the code changes and see if it makes sense and seems reasonable.
- [x] Optional: If you have a working local setup, you could try indexing gale pages and see if you get any errors ( 500 or otherwise): `python manage.py index_pages --gale`

### DIFF
--- a/ppa/archive/gale.py
+++ b/ppa/archive/gale.py
@@ -162,7 +162,7 @@ class GaleAPI:
         ]:
             # occasionally we get a 500 error when indexing all pages
             # refreshing API key and trying again, but log the error
-            if resp.status_code == requests.code.server_error:
+            if resp.status_code == requests.codes.server_error:
                 # Not sure yet if response has any meaningful content
                 logger.error(f"500 error on {rqst_url}: {resp.content}")
 


tanhaow:
I've reviewed the code changes and they look good to me:
- The approach of treating 500 errors similar to 401 errors is sensible and reuses existing retry logic.
- The unit test coverage appropriately verifies the new behavior.

I ran the test command (python manage.py index_pages --gale) on my local machine, and no 500 errors were encountered this time - All Gale API calls returned 200 OK responses.
But I did encounter 500 errors yesterday when I first set up my GALE_API_USERNAME configuration (so I gave up on retriving full-text data and just worked with work-level data), but I wasn't able to reproduce that issue today. 
The 500 errors may be intermittent/timing-related, so the retry logic is sensible.

tanhaow (approved):
I've reviewed the code changes and they look good to me:
- The approach of treating 500 errors similar to 401 errors is sensible and reuses existing retry logic.
- The unit test coverage appropriately verifies the new behavior.

I ran the test command (python manage.py index_pages --gale) on my local machine, and no 500 errors were encountered this time - All Gale API calls returned 200 OK responses.
But I did encounter 500 errors yesterday when I first set up my GALE_API_USERNAME configuration (so I gave up on retriving full-text data and just worked with work-level data), but I wasn't able to reproduce that issue today. 
The 500 errors may be intermittent/timing-related, so the retry logic is sensible.

PR #761, opened 2025-07-24
minor fix for Zotero and index logging

tanhaow:
Changes:
- Remove Solr dependency for COinS metadata generation on digitized work detail pages.
- Simplify and enhance `urlencode` logic.
- Update the logger to keep only the status code, URL, and item ID. #754 
- Fix minor formatting issue in CHANGELOG.rst.

Anything else need to be improved?

### DIFF
--- a/ppa/archive/gale.py
+++ b/ppa/archive/gale.py
@@ -79,6 +79,9 @@ class GaleAPI:
     #: base URL for all API requests
     api_root = "https://api.gale.com/api"
 
+    #: maximum number of retry attempts for API requests
+    max_retries = 3
+
     #: shared singleton instance; populated on first instantiation
     instance = None
 
@@ -142,11 +145,15 @@ class GaleAPI:
             rqst_opts["params"]["api_key"] = self.api_key
 
         resp = self.session.get(rqst_url, stream=stream, **rqst_opts)
-        logger.debug(
-            "get %s %s: %f sec",
+        # Log request - use info level for retries, debug for initial attempts
+        log_level = logger.info if retry > 0 else logger.debug
+        retry_info = f" (retry {retry}/{self.max_retries})" if retry > 0 else ""
+        log_level(
+            "get %s %s: %f sec%s",
             rqst_url,
             resp.status_code,
             resp.elapsed.total_seconds(),
+            retry_info,
         )
         if resp.status_code == requests.codes.ok:
             return resp
@@ -160,32 +167,18 @@ class GaleAPI:
             requests.codes.unauthorized,
             requests.codes.server_error,
         ]:
-            # Extract item ID from the URL (format: .../v1/item/GALE%7C{item_id})
-            item_id = None
-            if "GALE%7C" in rqst_url:
-                item_id = rqst_url.split("GALE%7C")[-1]
-
             # occasionally we get a 500 error when indexing all pages
             # refreshing API key and trying again, but log the error
             if resp.status_code == requests.codes.server_error:
-                # Log concise error info (status, URL, item ID)
-                logger.error(
-                    f"500 server error on {rqst_url}, Item ID: {item_id or 'N/A'}"
-                )
+                # Log concise error info (status, URL)
+                logger.error(f"500 server error on {rqst_url}")
 
             # If we get a 401 or 500 on a request that requires an api key,
-            # get a fresh key and then try the same request again (up to 3 retry attempts)
-            if requires_api_key and retry < 3:
-                # Log retry attempt for debugging - helps track API key refresh success
-                current_attempt = retry + 1
-                logger.info(
-                    f"Retrying after refreshing API key (attempt {current_attempt}/3) "
-                    f"- Status: {resp.status_code}, Item ID: {item_id or 'N/A'}"
-                )
-
+            # get a fresh key and then try the same request again (up to max_retries attempts)
+            if requires_api_key and retry < self.max_retries:
                 self.refresh_api_key()
 
-                retry_response = self._make_request(
+                return self._make_request(
                     url,
                     params=params,
                     requires_api_key=requires_api_key,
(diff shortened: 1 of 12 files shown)


ppa/archive/gale.py
```diff
@@ -160,23 +160,57 @@ def _make_request(
             requests.codes.unauthorized,
             requests.codes.server_error,
         ]:
+            # Extract item ID from the URL (format: .../v1/item/GALE%7C{item_id})
+            item_id = None
+            if "GALE%7C" in rqst_url:
+                item_id = rqst_url.split("GALE%7C")[-1]
+
```
rlskoeser, line 167: Ah, I forgot that we don't have an item id in the context where we're making the actual request. I don't think we should have this logic here; let's drop this and remove the item id from the logging. The URL is enough to see what's going on.

Sorry for misleading you.

ppa/archive/gale.py
```diff
@@ -160,23 +160,57 @@ def _make_request(
...
+                # Log concise error info (status, URL, item ID)
+                logger.error(
+                    f"500 server error on {rqst_url}, Item ID: {item_id or 'N/A'}"
+                )
 
             # If we get a 401 or 500 on a request that requires an api key,
-            # get a fresh key and then try the same request again
-            if requires_api_key and retry < 1:
+            # get a fresh key and then try the same request again (up to 3 retry attempts)
+            if requires_api_key and retry < 3:
```
rlskoeser, line 178: I had this set to retry once because when I initially implemented, that was all that was needed - when the API key expires, you get an error, and then when you refresh it and it works. I don't know what the behavior is with the 500 error or if it's consistent enough for us to test.

That said, I like this addition. Suggest moving it to a class variable, something like `max_retries` so it's easy to find and change if we want to adjust.

ppa/archive/gale.py
```diff
@@ -160,23 +160,57 @@ def _make_request(
...
                     requires_api_key=requires_api_key,
                     stream=stream,
                     retry=retry + 1,
                 )
+
+                # Log successful retry - confirms API key refresh worked and API call succeeded
+                logger.info(
+                    f"Retry succeeded with HTTP {retry_response.status_code} "
+                    f"(attempt {current_attempt}/3) - Item ID: {item_id or 'N/A'}"
+                )
```
rlskoeser, line 200: I like this thought but I don't think we need it - we'll end up with redundant logging, since we already have debug logging for every request (~ line 145).

What if we include retry in the earlier logging if it's non-zero? And maybe also elevate from debug to info for retries.

ppa/archive/gale.py
```diff
@@ -160,23 +160,57 @@ def _make_request(
...
+            else:
+                # Log when we decide not to retry (doesn't require API key or max retries)
+                if requires_api_key and retry >= 3:
+                    logger.warning(
+                        f"Not retrying - 3 retries exhausted, Item ID: {item_id or 'N/A'}"
+                    )
+                elif not requires_api_key:
+                    logger.warning(
+                        f"Not retrying - no API key required, Item ID: {item_id or 'N/A'}"
+                    )
```
rlskoeser, line 212: This is great - once we have this on staging it should be obvious from the logs whether or not we recover from the 500 errors.

ppa/archive/models.py
```diff
@@ -584,6 +584,11 @@ def has_fulltext(self):
         HathiTrust, Gale, or EEBO-TCP)."""
         return self.source in [self.HATHI, self.GALE, self.EEBO]
 
+    @property
+    def work_type(self):
+        """Work type formatted for COinS metadata (matches Solr field format)."""
+        return self.get_item_type_display().lower().replace(" ", "-")
```
rlskoeser, line 590: Good solution. Please update the `index_data` method to use this property (~ line 1009) so we can guarantee they stay in sync.

ppa/archive/templatetags/ppa_tags.py
```diff
@@ -129,7 +129,15 @@ def _get_item_value(obj, key, default=None):
...
-            return value
+            # If it's a method (Django model method), call it
+            if callable(value):
+                try:
+                    return value()
+                except TypeError:
+                    # Method requires arguments, skip it
+                    pass
+            else:
+                return value
```
rlskoeser, line 140: Is this for the `first_page`  method on `DigitizedWork` ? What do you think about changing that method to a property (or cached property) instead? (Seems cleaner and simpler to me to fix it there - I don't remember now why I didn't make it a property before. )

rlskoeser (changes requested):
Revisions look really good.  Thank you for fixing my formatting mistakes on the change log file.

I'm curious, were you able to get the Gale API to hit the retry logic at all in testing?

I'm requesting a few small changes:
- drop item_id from Gale API logging (my fault, sorry for misleading you)
- drop redundant logging for Gale API retries
- make max retries a class variable on the Gale API
- use new `work_type` property in `index_data` method
- convert `first_page` (and any other methods) to property or cached property instead of adding callable logic to the coins template code

ppa/archive/gale.py
tanhaow: Great suggestion.

ppa/archive/templatetags/ppa_tags.py
tanhaow: You're so smart!!

tanhaow:
> I'm curious, were you able to get the Gale API to hit the retry logic at all in testing?

To the question -- No, unfortunately. These unit tests can only verify the retry logic works correctly with fake responses, but they cannot test the real API. To actually test the retry logic, it seems like we have to run it on staging.

I've made the following changes:
- [ ] drop item_id from Gale API logging
- [ ] drop redundant logging for Gale API retries
- [ ] make max retries a class variable on the Gale API
- [ ] use new work_type property in index_data method
- [ ] convert first_page (and any other methods) to property or cached property instead of adding callable logic to the coins template code

Let me know if you have any more suggestions - You're such a smart person!!

PR #792, opened 2026-03-23
Implement Basic Authentication for Gale Content API

### LINKED ISSUE #790: Update Gale API code to support basic auth
Gale is switching their content API to Basic Auth for security reasons, so we need to update our code to handle this change. We're already using `requests` so it shouldn't be difficult.  In addition to the code change in the Gale API class, we'll need some configuration changes. We'll need add a django setting for the client secret (local settings), and the credentials should be added to cdh-ansible as a vaulted variable which is included in the local settings on deploy.

tanhaow:
**Associated Issue(s):** resolves #790

**Note:** This PR replaces the closed PR #791. The branch history was rewritten using `git-filter-repo` to [remove accidentally committed credentials](https://github.com/Princeton-CDH/ppa-django/commit/b8e0a2d2d91f9f69a128d425da6ac96e538166db). All commits now use fake test credentials.

### Changes in this PR

- AImplement Basic Auth credentials settings (`GALE_API_LOCATION_ID` and `GALE_API_SECRET`) for Gale Content API requests
- Update local_settings.py.sample with new required configuration settings

### Notes

- Basic Authorization header is generated using `Base64` encoding of `location_id;secret` format
- Found and fix singleton re-initialization bug: Added a re-initialization guard in `gake.py` to prevent session loss. This is to ensure the Authorization header persists across multiple `GaleAPI()` instantiations.
- Sphinx documentation build failure is unrelated to this PR (pre-existing issue with Waffle flag initialization during docs build)

### Reviewer Checklist

- [x] Verify Basic Auth header encoding matches Gale specification
- [x] Review the code changes

### DIFF
--- a/DEPLOYNOTES.rst
+++ b/DEPLOYNOTES.rst
@@ -3,6 +3,15 @@
 Deploy and Upgrade notes
 ========================
 
+3.17
+----
+
+* Gale Content API now requires Basic Auth credentials for API access.
+  **GALE_API_LOCATION_ID** and **GALE_API_SECRET** must be configured in
+  local settings with the credentials provided by Gale. The existing
+  **GALE_API_USERNAME** setting is still required for API key generation.
+  Both authentication mechanisms (Basic Auth and API key) work together.
+
 3.16
 ----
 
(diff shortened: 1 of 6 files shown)


rlskoeser (commented):
Let's not hold this up any longer waiting to hear back on the implementation. 

Before merging, please add a comment explaining why we're not using built-in basic auth and would you also clean up an old print statement?

rlskoeser (approved):

PR #794, opened 2026-06-17
Release/3.16.1

tanhaow:
**Associated Issue(s):** resolves #793 

### Changes in this PR

- Add Basic Auth support to Gale Content API (`GALE_API_SECRET` setting required)
- Update changelog for 3.17 release
- Update `requirements.lock` and npm dependencies

### Notes

- The PR diff appears to only show meta file changes due to a git history issue: the feature branch was accidentally merged directly into `main` then reverted. The code changes are present in this release branch — run `git diff main release/3.17 -- ppa/archive/gale.py` to verify.

### Reviewer Checklist

- [ ] Confirm Gale Basic Auth code is present after merge


CHANGELOG.rst
```diff
@@ -3,6 +3,11 @@
 CHANGELOG
 =========
 
+3.17
```
rlskoeser: How do you feel about making this a micro release? It's not a bugfix exactly but it's not a new feature.

rlskoeser (changes requested):
Where are the code changes? I only see meta package files (changelog, readme, and dependency lock files).  Shouldn't we be seeing updates to the Gale API code?

Please update the PR description.

CHANGELOG.rst
tanhaow: Yeah, I had the same feeling that it's a bit big for what it is. I went with 3.17 because it didn't feel like a clean either, but I'm happy to bump it down to 3.16.1 if you think that fits better.

rlskoeser: yes, let's call it 3.16.1.

CHANGELOG.rst
```diff
@@ -3,6 +3,11 @@
...
+----
+
+- Update Gale Content API integration to use Basic Auth in addition to API key authentication
```
rlskoeser:
```suggestion
- Update DOI in the readme to use the DOI for all versions of the code
```

rlskoeser (approved):
👍

CHANGELOG.rst
tanhaow: Just noticed the suggested edit replaced the Gale API entry rather than adding to it, so the tag originally created by git flow was missing it. I've manually fixed this.

